PR #8, opened 2016-12-15
Add documentation generation.

cmnbroad:
- There are 2 commits; the first is a temporary diff commit for code review purposes only.
- The documentation generation is dependent on and tied to the (new) CommandLineArgumentParser, and has some knowledge of internal CommandLineArgumentParser.ArgumentDefinition class. I didn't factor that class out as part of this PR, but it should be done separately.
- An example of the documentation generated by running this code on a version of GATK4 modified to use the DocumentedFeature annotation can be seen at this (internal Broad internal) location: ~cnorman/gatkDoc/index.html.
- There is a wiki page under construction [here](https://github.com/broadinstitute/barclay/wiki/Using-the-Documentation-Generation-utilities-in-Barclay) that shows how to employ these classes to generate custom documentation.


cmnbroad:
@vdauwera For review. @droazen will also review; let me know if you have any questions.

src/main/java/org/broadinstitute/barclay/argparser/Argument.java
```diff
@@ -97,9 +97,46 @@
...
     boolean overridable() default false;
 
+    /**
+     * Hard lower bound on the allowed value for the annotated argument -- generates an exception if violated.
+     * Enforced only for numeric types whose values are explicitly specified on the command line.
+     *
+     * @return Hard lower bound on the allowed value for the annotated argument, or Double.NEGATIVE_INFINITY
+     *         if there is none.
+     */
+    double minValue() default Double.NEGATIVE_INFINITY;
```
magicDGS: Are these bounds used in the parsing? If so, I would like to have this soon!! Thanks for it!

src/main/java/org/broadinstitute/barclay/argparser/CommandLineProgramProperties.java
```diff
@@ -27,5 +26,6 @@
      */
     String usageExample() default "The author of this program hasn't included any example usage, please complain to them.";
```
vdauwera: Would be great (future feature) to be able to emit a list of usage-lacking tools at build time.

src/main/resources/helpTemplates/generic.template.html
```diff
@@ -0,0 +1,215 @@
...
+					</#list>
+				</ul>
+			</#if>
+			<#if extradocs?size != 0 || arguments.all?size != 0>
+				<hr>
+				<h2>Command-line Arguments</h2>
+				<p></p>
+			</#if>
+			<#if extradocs?size != 0>
+				<h3>Engine arguments</h3>
```
vdauwera: Is the "extra docs"/"engine arguments" thing still applicable? Based on our discussion yesterday I thought that engine args are now interspersed among the other tool args (or grouped as shared args if tagged accordingly).

src/main/resources/helpTemplates/common.html
```diff
@@ -0,0 +1,53 @@
+<#--
+        This file contains part of the theming used to present Barclay docs on a website. Styling is separated
+        out, so pages will be minimalistic html unless replacement styling is provided.
+        -->
+
+    <#macro footerInfo>
+        <hr>
+        <p><a href='#top'><i class='fa fa-chevron-up'></i> Return to top</a></p>
```
vdauwera: FYI some elements here are very specific to libraries we use in the GATK website, like `i class='fa fa-...'` and `class="nav nav-pills nav-stacked" id="sidenav"`. I don't mind since it's what we use, but if you're trying to be as generic as possible, would you want to move this out entirely? I could propose a completely generic template if that helps.

src/main/resources/helpTemplates/generic.index.template.html
```diff
@@ -0,0 +1,67 @@
...
+?>
+
+<div class='row-fluid'>
+
+<div class='span9'>
+
+<#include "common.html"/>
+
+<#macro emitGroup group>
+    <div class="accordion-group">
```
vdauwera: Same comment as previous regarding library-specific formatting code.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineProgramProperties.java
cmnbroad: @vdauwera I'm glad you caught this. This came from GATK4 (and there are only a handful of tools for which is populated), but it was never integrated with the CLP, and I didn't integrate it with docgen since it wasn't present in the GATK3 code base. So we need to figure out how/where it should surface. Again I think I'll make a separate ticket for that so as not to hold up the current PRs.

Also, BTW, Picard doesn't currently have this either; it looks like the usage summary there is included by convention in the "summary" output. It would be nice if they were separated since those strings are currently Picard-specific (i.e., they have embedded "java -jar picard.jar" references that won't make sense if displayed on the context of GATK).

src/main/resources/helpTemplates/generic.template.html
cmnbroad: You're right that we don't need this for engine args since they're determined by the "common" property, but I left this in the code since it seemed like a nice general mechanism for referencing additional docs (was this previously used exclusively for engine arguments ? it looks that way based on how it appears in the template.) However, the "Engine arguments" bit should be removed from the template here).

src/main/resources/helpTemplates/common.html
cmnbroad: @vdauwera Once thing I tried to do is to keep Barclay so that it could generate some minimally useful doc by default with minimal modifications to the template. I tried to remove anything that was truly GATK specific, but I don't mind leaving these styles in here since they make the output nicer.

src/main/java/org/broadinstitute/barclay/argparser/Argument.java
vdauwera: Splendid, thank you

src/main/java/org/broadinstitute/barclay/argparser/CommandLineProgramProperties.java
vdauwera: That's right, historically both GATK3 and Picard have only included a usage example as part as the general tool description (if at all). I love the idea of providing a usage example through an explicit declaration that can be rendered in the appropriate syntax. 

I agree it can be a followup ticket too.

src/main/resources/helpTemplates/generic.template.html
vdauwera: AFAIK we've only used it for engine args, yep. I agree it's a mechanism that has useful potential even if we don't have a specific use case right off the bat. If we're keeping it for now let's just strip off the excessively specific bits.

src/main/resources/helpTemplates/common.html
vdauwera: If/when I find some time I'll make a PR to make the default output more generic, but I agree it's not a problem to keep this stuff in for now.

droazen:
I was going to review the plugin aspect of this branch (eg., read filter handling), but for a merge into barclay I think @vdauwera's review is sufficient, so I've untagged myself as a reviewer. Instead, I'll see what the plugin support looks like from the client perspective in GATK, and give my feedback at that time.

@cmnbroad, merge as soon as @vdauwera gives the all-clear.

vdauwera:
There were a couple of todos that we discussed in person on Monday, that involved removing the full list of read filters from individual tool docs, yet exposing arguments of the tool's default read filters.

droazen:
@vdauwera Perhaps those would be better done in a followup PR?

vdauwera:
Happy to leave that up to @cmnbroad's judgment

cmnbroad:
@droazen @vdauwera I'm trying to see if I can get most of them done now, but I think at least one will have to be done separately, since I don't want to delay merging this past tomorrow.

vdauwera:
Feel free to merge when you feel comfortable with the state of things, @cmnbroad. I also would prefer this to go in before the break, as long as there are clear tickets for the followup work that we can pick up in the new year.

cmnbroad:
- Removed the allowed values inline links for each plugin descriptor (i.e. read filter), so the output is effectively the same as GTAK3 now. I’ve created a separate [ticket](https://github.com/broadinstitute/barclay/issues/12) to add an inline link to the read filter group, but this will be a separate ticket/PR. 
- Removed the dependent args from the args list (I’ve left the code for generating these in Barclay, and will just remove the references to them from the GATK template).
- Created a [ticket](https://github.com/broadinstitute/barclay/issues/13) to add a feature to list the dependent arguments only for default plugin instances (i.e., for each default readfilter listed for a given tool, include the filter's dependent arguments somewhere).
- There is a separate ticket to do something with the currently unused usageSummary property https://github.com/broadinstitute/barclay/issues/10
- Did some code cleanup

Issue #6, opened 2016-12-14
@Hidden, @Advanced annotations are not handled by command line parser


magicDGS:
I don't see any handling of annotations for arguments:

* `@Hidden`: I expect that it hides the option from help, but not for the argument parsing.
* `@Advance`: I expect this option to appear under an especial category pointing out that they should be used with caution. I think that another interesting way for this advance options could be an argument in `SpecialArgumentsCollection` to show the "advance" help.

cmnbroad:
@magcDGS You're right - Hidden and Advanced should both be handled specially in the command line usage output.

cmnbroad:
We should also integrate min/max and recommended values as well (assuming they're integrated as part of the docgen changes to @Argument).

cmnbroad:
These annotations and properties all need to be reflected/honored in both parsing and usage output.

cmnbroad:
Partially fixed (for max/min values) via https://github.com/broadinstitute/barclay/pull/11. Still need to do the Hidden/Advanced part.

PR #11, opened 2016-12-21
Handling max/min values in CommandLineArgumentParser

magicDGS:
### Included
* Handling max/min values in `CommandLineArgumentParser` (solves partially #6)
* New `CommandLineArgumentException` for out-of-range values.

### Not included
* `LegacyCommandLineArgumentParser` does not have max/min validation: Picard is not supporting currently this feature in its parser.
* Recommended boundaries are not handled yet.

### DIFF
--- a/src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
+++ b/src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
@@ -8,6 +8,8 @@ import joptsimple.OptionSpecBuilder;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.text.WordUtils;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.broadinstitute.barclay.utils.Utils;
 
 import java.io.BufferedReader;
@@ -27,6 +29,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.DoubleFunction;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -69,6 +72,8 @@ public final class CommandLineArgumentParser implements CommandLineParser {
     private static final String NULL_STRING = "null";
     public static final String COMMENT = "#";
     public static final String POSITIONAL_ARGUMENTS_NAME = "Positional Argument";
+
+    private static final Logger logger = LogManager.getLogger();
 
     // Map from (full class) name of each CommandLinePluginDescriptor requested and
     // found to the actual descriptor instance
@@ -508,6 +513,52 @@ public final class CommandLineArgumentParser implements CommandLineParser {
         // finally, give each plugin a chance to trim down any unseen instances from it's own list
         pluginDescriptors.entrySet().forEach(e -> e.getValue().validateArguments());
     }
+    /**
+     * Check the provided value against any range constraints specified in the Argument annotation
+     * for the corresponding field. Throw an exception if limits are violated.
+     *
+     * - Only checks numeric types (int, double, etc.)
+     */
+    private void checkArgumentRange(final ArgumentDefinition argumentDefinition, final Object argumentValue) {
+        // Only validate numeric types because we have already ensured at constructor time that only numeric types have bounds
+        if (!Number.class.isAssignableFrom(argumentDefinition.type)) {
+            return;
+        }
+
+        final Double argumentDoubleValue = (argumentValue == null) ? null : ((Number)argumentValue).doubleValue();
+
+        // Check hard limits first, if specified
+        if (isOutOfRange(argumentDefinition.minValue, argumentDefinition.maxValue, argumentDoubleValue)) {
+            throw new CommandLineException.OutOfRangeArgumentValue(argumentDefinition.getLongName(), argumentDefinition.minValue, argumentDefinition.maxValue, argumentValue);
+        }
+        // Check recommended values
+        if (isOutOfRange(argumentDefinition.minRecommendedValue, argumentDefinition.maxRecommendedValue, argumentDoubleValue)) {
+            final boolean outMinValue = argumentDefinition.minRecommendedValue != Double.NEGATIVE_INFINITY;
+            final boolean outMaxValue = argumentDefinition.maxRecommendedValue != Double.POSITIVE_INFINITY;
+            if (outMinValue && outMaxValue) {
+                logger.warn("Argument --{} has value {}, but recommended within range ({},{})",
+                        argumentDefinition.getLongName(), argumentDoubleValue, argumentDefinition.minRecommendedValue, argumentDefinition.maxRecommendedValue);
+            } else if (outMinValue) {
+                logger.warn("Argument --{} has value {}, but minimum recommended is {}",
+                        argumentDefinition.getLongName(), argumentDoubleValue, argumentDefinition.minRecommendedValue);
+            } else if (outMaxValue) {
+                logger.warn("Argument --{} has value {}, but maximum recommended is {}",
+                        argumentDefinition.getLongName(), argumentDoubleValue, argumentDefinition.maxRecommendedValue);
+            }
+        }
+    }
+
+    // null values are always out of range
+    private static boolean isOutOfRange(final double minValue, final double maxValue, final Double value) {
+        return value == null || minValue != Double.NEGATIVE_INFINITY && value < minValue
+                || maxValue != Double.POSITIVE_INFINITY && value > maxValue;
+    }
+
+    // check if the value is infinity or a mathematical integer
+    private static boolean isInfinityOrMathematicalInteger(final double value) {
+        return Double.isInfinite(value) || value == Math.rint(value);
+    }
+
 
     @SuppressWarnings("unchecked")
     private void setPositionalArgument(final String stringValue) {
@@ -566,6 +617,9 @@ public final class CommandLineArgumentParser implements CommandLineParser {
             } else {
                 value = constructFromString(CommandLineParser.getUnderlyingType(argumentDefinition.field), stringValue, argumentDefinition.getLongName());
             }
+
+            // check the argument range
+            checkArgumentRange(argumentDefinition, value);
 
             if (argumentDefinition.isCollection) {
                 @SuppressWarnings("rawtypes")
@@ -947,6 +1001,7 @@ public final class CommandLineArgumentParser implements CommandLineParser {
 
     public static class ArgumentDefinition {
         public final Field field;
+        public final Class<?> type;
         final String fieldName;
         public final String fullName;
         public final String shortName;
@@ -961,6 +1016,10 @@ public final class CommandLineArgumentParser implements CommandLineParser {
         final boolean isSpecial;
         final boolean isSensitive;
         public final CommandLinePluginDescriptor<?> controllingDescriptor;
+        final Double maxValue;
+        final Double minValue;
+        final Double maxRecommendedValue;
+        final Double minRecommendedValue;
 
         public ArgumentDefinition(
                 final Field field,
@@ -998,8 +1057,30 @@ public final class CommandLineArgumentParser implements CommandLineParser {
             //null collections have been initialized by createCollection which is called in handleArgumentAnnotation
             //this is optional if it's specified as being optional or if there is a default value specified
             this.optional = annotation.optional() || ! this.defaultValue.equals(NULL_STRING);
+            this.maxValue = annotation.maxValue();
+            this.minValue = annotation.minValue();
+            this.maxRecommendedValue = annotation.maxRecommendedValue();
+            this.minRecommendedValue = annotation.minRecommendedValue();
+            // bounds should be only set for numeric arguments and if the type is integer it should
+            // be set to an integer
+            this.type = CommandLineParser.getUnderlyingType(this.field);
+            if (! Number.class.isAssignableFrom(this.type)) {
+                if (this.maxValue != Double.POSITIVE_INFINITY
+                        || this.minValue != Double.NEGATIVE_INFINITY
+                        || this.maxRecommendedValue != Double.POSITIVE_INFINITY
+                        || this.minRecommendedValue != Double.NEGATIVE_INFINITY) {
+                    throw new CommandLineException.CommandLineParserInternalException(String.format("Min/max value ranges can only be set for numeric arguments. Argument --%s has a minimum or maximum value but has a non-numeric type.", this.getLongName()));
+                }
+            }
+            if (Integer.class.isAssignableFrom(this.type)) {
+                if (!isInfinityOrMathematicalInteger(this.maxValue)
+                        || !isInfinityOrMathematicalInteger(this.minValue)
+                        || !isInfinityOrMathematicalInteger(this.maxRecommendedValue)
+                        || !isInfinityOrMathematicalInteger(this.minRecommendedValue)) {
+                    throw new CommandLineException.CommandLineParserInternalException(String.format("Integer argument --%s has a minimum or maximum attribute with a non-integral value.", this.getLongName()));
+                }
+            }
         }
-
 
         public Object getFieldValue(){
             try {
--- a/src/main/java/org/broadinstitute/barclay/argparser/CommandLineException.java
+++ b/src/main/java/org/broadinstitute/barclay/argparser/CommandLineException.java
@@ -1,5 +1,7 @@
 package org.broadinstitute.barclay.argparser;
 
+import java.util.function.DoubleFunction;
+
 /**
  * Exceptions thrown by CommandLineParser implementations.
  */
@@ -37,6 +39,36 @@ public class CommandLineException extends RuntimeException {
         public BadArgumentValue(String message) {
             super(String.format("Illegal argument value: %s", message));
         }
+    }
+
+    public static class OutOfRangeArgumentValue extends BadArgumentValue {
+        private static final long serialVersionUID = 0L;
+
+        public OutOfRangeArgumentValue(final String argName, final double minValue, final double maxValue, final Object value) {
+            super(argName, getValueString(value), getMessage(minValue, maxValue, value instanceof Integer));
+        }
+
+        // to handle null values
+        private static String getValueString(final Object value) {
+            return (value == null) ? "null" : value.toString();
+        }
+
+        // get the message for the values, correctly formatted
+        private static String getMessage(final double minValue, final double maxValue, final boolean asInt) {
+            final boolean outMinValue = minValue != Double.NEGATIVE_INFINITY;
+            final boolean outMaxValue = maxValue != Double.POSITIVE_INFINITY;
+            final DoubleFunction<String> toString = (asInt) ? Double::toString : v -> Integer.toString((int) Math.rint(v));
+            if (outMinValue && outMaxValue) {
+                return String.format("allowed range [%s, %s].", toString.apply(minValue), toString.apply(maxValue));
+            } else if (outMinValue) {
+                return String.format("minimum allowed value %s", toString.apply(minValue));
+            } else if (outMaxValue) {
+                return String.format("maximum allowed value %s", toString.apply(maxValue));
+            }
+            // this should never be reached
+            throw new IllegalArgumentException("Unbounded range should not thrown this exception");
+        }
+
     }
 
     /**
--- a/src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
+++ b/src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
@@ -24,6 +24,8 @@
 package org.broadinstitute.barclay.argparser;
 
 import org.apache.commons.lang3.text.WordUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 import java.io.BufferedReader;
 import java.io.File;
@@ -98,6 +100,8 @@ public class LegacyCommandLineArgumentParser implements CommandLineParser {
     };
 
     private final Set<String> optionsThatCannotBeOverridden = new HashSet<String>();
+
+    private final static Logger logger = LogManager.getLogger();
 
     /**
      * A typical command line program will call this to get the beginning of the usage message,
@@ -548,7 +552,6 @@ public class LegacyCommandLineArgumentParser implements CommandLineParser {
             } else {
                 value = constructFromString(getUnderlyingType(optionDefinition.field), stringValue);
             }
-
         } catch (final CommandLineException e) {
             messageStream.println("ERROR: " + e.getMessage());
             return false;
@@ -806,7 +809,6 @@ public class LegacyCommandLineArgumentParser implements CommandLineParser {
                 printOrder = fieldPosition * 1000;
             }
 
-
             final OptionDefinition optionDefinition = new OptionDefinition(field,
                     field.getName(),
                     optionAnnotation.shortName(),
@@ -815,6 +817,20 @@ public class LegacyCommandLineArgumentParser implements CommandLineParser {
                     optionAnnotation.maxElements(), field.get(callerOptions), optionAnnotation.common(),
                     optionAnnotation.mutex(),
                     printOrder);
+
+            // log a warning if boundaries are set
+            if (optionAnnotation.maxValue() != Double.POSITIVE_INFINITY) {
+                logger.warn("Maximum allowed value for argument --{} is not enforced", optionDefinition.name);
+            }
+            if (optionAnnotation.minValue() != Double.NEGATIVE_INFINITY) {
+                logger.warn("Minimum allowed value for argument --{} is not enforced", optionDefinition.name);
+            }
+            if (optionAnnotation.maxRecommendedValue() != Double.POSITIVE_INFINITY) {
+                logger.warn("Maximum recommended value for argument --{} is not checked", optionDefinition.name);
+            }
+            if (optionAnnotation.minRecommendedValue() != Double.NEGATIVE_INFINITY) {
+                logger.warn("Minimum recommended value for argument --{} is not checked", optionDefinition.name);
+            }
 
             for (final String option : optionAnnotation.mutex()) {
                 final OptionDefinition mutextOptionDef = optionMap.get(option);
--- a/src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
+++ b/src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
@@ -749,6 +749,88 @@ public final class CommandLineArgumentParserTest {
         return out.toString();
     }
 
+    @Test(expectedExceptions = CommandLineException.CommandLineParserInternalException.class)
+    public void testWithBoundariesArgumentsForNoNumeric() {
+        @CommandLineProgramProperties(summary = "broken tool",
+                oneLineSummary = "broken tool",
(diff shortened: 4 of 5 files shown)


magicDGS:
This is very simple but useful for my work. Could you have a look to this, @cmnbroad?

src/main/java/org/broadinstitute/barclay/argparser/CommandLineException.java
```diff
@@ -39,6 +39,28 @@ public BadArgumentValue(String message) {
...
+
+        public OutOfRangeArgumentValue(final String arg, final double minValue, final double maxValue, final double value) {
+            super(arg, Double.toString(value), getMessage(minValue, maxValue));
+        }
+
+        private static String getMessage(final double minValue, final double maxValue) {
+            final boolean outMinValue = minValue != Double.NEGATIVE_INFINITY;
+            final boolean outMaxValue = maxValue != Double.POSITIVE_INFINITY;
+            if (outMinValue && outMaxValue) {
+                return String.format("allowed range (%.2f, %.2f).", minValue, maxValue);
```
cmnbroad: I would widen the precision when converting these to strings, otherwise the values may be unrecognizable. Maybe .5f is reasonable ? Or maybe Double.toString will do a better job ?

src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
```diff
@@ -548,7 +548,8 @@ private boolean parseOption(String key, final String stringValue, final boolean
             } else {
                 value = constructFromString(getUnderlyingType(optionDefinition.field), stringValue);
             }
-
+            // TODO - Picard legacy parsing is not taking into account range within options
+            // TODO - should we add support here?
```
cmnbroad: I don't think we want to invest in adding support for these to the legacy parser. Perhaps just add some code to log a warning saying its unsupported and not enforced if a value is specified. (We don't want to throw since the new parser will handle them, and we want to be able to switch back and forth between the two parsers).

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -498,6 +498,26 @@ private void validatePluginArguments() {
...
+            return;
+        }
+        final double argumentDoubleValue = ((Number)argumentValue).doubleValue();
+
+        // Check hard limits first, if specified
+        if ((argumentDefinition.minValue != Double.NEGATIVE_INFINITY && argumentDoubleValue < argumentDefinition.minValue)
+            || (argumentDefinition.maxValue != Double.POSITIVE_INFINITY && argumentDoubleValue > argumentDefinition.maxValue)) {
+            throw new CommandLineException.OutOfRangeArgumentValue(argumentDefinition.fullName, argumentDefinition.minValue, argumentDefinition.maxValue, argumentDoubleValue);
+        }
+    }
```
cmnbroad: While we're doing this, we might want to log a warning if the value is outside the minRec/maxRec range as well (you didn't include those annotation fields in this PR, but when you rebase this they'll be there!).

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -498,6 +498,26 @@ private void validatePluginArguments() {
...
     }
+    /**
+     * Check the provided value against any range constraints specified in the Argument annotation
+     * for the corresponding field. Throw an exception if limits are violated.
+     *
+     * - Only checks numeric types (int, double, etc.)
+     */
+    private void checkArgumentRange(final ArgumentDefinition argumentDefinition, final Object argumentValue) {
+        // Only validate numeric types
+        if ( ! (argumentValue instanceof Number) ) {
```
cmnbroad: argumentValue can be null if the user specifies "null". This code won't issue a validation failure for that case, but it probably should.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
```diff
@@ -749,6 +749,56 @@ private static String captureSystemStream(Runnable runnable, PrintStream stream,
...
+            summary = "tool with boundaries",
+            oneLineSummary = "tools with boundaries",
+            programGroup = TestProgramGroup.class
+    )
+    public class WithBoundariesArguments {
+
+        @Argument(doc = "Integer in the range [0, 30]", optional = true, minValue = 0, maxValue = 30)
+        public int integerArg = 20;
+
+        @Argument(doc = "String argument", optional = true, minValue = 0, maxValue = 30)
```
cmnbroad: Per my comment above, I think this case should throw, since the range is not honored/enforced. We should keep a test for this though to make sure it fails as we think it should.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineException.java
```diff
@@ -39,6 +39,28 @@ public BadArgumentValue(String message) {
...
+            } else if (outMinValue) {
+                return String.format("minimum allowed value %.2f", minValue);
+            } else if (outMaxValue) {
+                return String.format("maximum allowed value %.2f", maxValue);
+            }
+            return "unkwnon";
```
cmnbroad: Spelling. "unknown"

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
```diff
@@ -749,6 +749,56 @@ private static String captureSystemStream(Runnable runnable, PrintStream stream,
...
+        Assert.assertEquals(o.stringArg, expectedString);
+    }
+
+    @DataProvider
+    public Object[][] outOfRangeArgs() {
+        return new Object[][]{
+                {new String[]{"--integerArg", "-45"}},
+                {new String[]{"--integerArg", "-1"}},
+                {new String[]{"--integerArg", "31"}},
+                {new String[]{"--integerArg", "106"}}
```
cmnbroad: Can you add a test that uses the value "null". Thx.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
```diff
@@ -749,6 +749,56 @@ private static String captureSystemStream(Runnable runnable, PrintStream stream,
...
 
+    @CommandLineProgramProperties(
+            summary = "tool with boundaries",
+            oneLineSummary = "tools with boundaries",
+            programGroup = TestProgramGroup.class
+    )
+    public class WithBoundariesArguments {
+
+        @Argument(doc = "Integer in the range [0, 30]", optional = true, minValue = 0, maxValue = 30)
+        public int integerArg = 20;
```
cmnbroad: Add a double argument test as well.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineException.java
cmnbroad: It would also be nice to handle the integer valued case correctly in the output strings (and not print out floating point values for an integer argument).

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -988,6 +1013,9 @@ public ArgumentDefinition(
             //null collections have been initialized by createCollection which is called in handleArgumentAnnotation
             //this is optional if it's specified as being optional or if there is a default value specified
             this.optional = annotation.optional() || ! this.defaultValue.equals(NULL_STRING);
+
+            this.maxValue = annotation.maxValue();
+            this.minValue = annotation.minValue();
```
cmnbroad: Also, it might be worth checking that the author hasn't provided a non-integral value for an integer arg.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
magicDGS: I added validation for null values, and here it only returns if the value is not numeric and not null.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
magicDGS: I rebased, added a logger in the class and log a warning if the value is out of the soft boundaries.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
magicDGS: I check both on `ArgumentDefinition`construction, for both hard and soft boundaries.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineException.java
```diff
@@ -39,6 +39,28 @@ public BadArgumentValue(String message) {
         }
     }
 
+    public static class OutOfRangeArgumentValue extends BadArgumentValue {
+        private static final long serialVersionUID = 0L;
+
+        public OutOfRangeArgumentValue(final String arg, final double minValue, final double maxValue, final double value) {
```
magicDGS: Done.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineException.java
magicDGS: I changed the signature of the constructor for take an `Object` and check if it is an `Integer` instance. Now for double it uses `Double.toString(value)` and `Integer.toString((int) Math.rint(value)`.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineException.java
magicDGS: Changed, although this point should never be reached. Maybe is better to thrown an `IllegalArgumentException` if so?

src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
magicDGS: Should we log the warning on `OptionDefinition` construction?

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
magicDGS: Done.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
magicDGS: Added test for throwing `CommandLineException.CommandLineParserInternalException`, and removed boundaries here.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
magicDGS: Added.

magicDGS:
I did some workarround for the implementation to handle everything in a better way. Thanks for reviewing @cmnbroad!

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -947,6 +1000,7 @@ private Object constructFromString(final Class clazz, final String s, final Stri
 
     public static class ArgumentDefinition {
         public final Field field;
+        public final Class<?> type;
```
magicDGS: This could be used in other parts of the code in this class instead of calling `CommandLineParser.getUnderlyingType()`. This will simplify the code a bit. Should I change it here, @cmnbroad?

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -998,9 +1056,31 @@ public ArgumentDefinition(
...
+                        || this.minRecommendedValue != Double.NEGATIVE_INFINITY) {
+                    throw new CommandLineException.CommandLineParserInternalException("Bounds for arguments could only be set for numeric arguments");
+                }
+            }
+            if (Integer.class.isAssignableFrom(this.type)) {
+                if (!isInfinityOrMathematicalInteger(this.maxValue)
+                        || !isInfinityOrMathematicalInteger(this.minValue)
+                        || !isInfinityOrMathematicalInteger(this.maxRecommendedValue)
+                        || !isInfinityOrMathematicalInteger(this.minRecommendedValue)) {
+                    throw new CommandLineException.CommandLineParserInternalException("Bounds for integer arguments should be mathematical integers");
```
cmnbroad: Same thing for this; test of the message should include the offending field name.

Also, please add a test for this code path.

src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
cmnbroad: Yes, sounds like a reasonable place (please include the name of the argument int he message here as well).

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -508,6 +513,51 @@ private void validatePluginArguments() {
...
+        // Check hard limits first, if specified
+        if (isOutOfRange(argumentDefinition.minValue, argumentDefinition.maxValue, argumentDoubleValue)) {
+            throw new CommandLineException.OutOfRangeArgumentValue(argumentDefinition.fullName, argumentDefinition.minValue, argumentDefinition.maxValue, argumentValue);
+        }
+        // Check recommended values
+        if (isOutOfRange(argumentDefinition.minRecommendedValue, argumentDefinition.maxRecommendedValue, argumentDoubleValue)) {
+            final boolean outMinValue = argumentDefinition.minRecommendedValue != Double.NEGATIVE_INFINITY;
+            final boolean outMaxValue = argumentDefinition.maxRecommendedValue != Double.POSITIVE_INFINITY;
+            if (outMinValue && outMaxValue) {
+                logger.warn("Argument --{} has value {}, but recommended within range ({},{})",
```
cmnbroad: We should have test cases that exercise these code paths (I don't think there are any ?), especially if we're going to use the {} style formatting, to at least ensure that we don't blow up here.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -508,6 +513,51 @@ private void validatePluginArguments() {
...
+                logger.warn("Argument --{} has value {}, but minimum recommended is {}",
+                        argumentDefinition.fullName, argumentDoubleValue, argumentDefinition.minRecommendedValue);
+            } else if (outMaxValue) {
+                logger.warn("Argument --{} has value {}, but maximum recommended is {}",
+                        argumentDefinition.fullName, argumentDoubleValue, argumentDefinition.maxRecommendedValue);
+            }
+        }
+    }
+
+    private static boolean isOutOfRange(final double minValue, final double maxValue, final Double value) {
```
cmnbroad: Add a comment here saying we consider a null value to be out of range.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -508,6 +513,51 @@ private void validatePluginArguments() {
...
         pluginDescriptors.entrySet().forEach(e -> e.getValue().validateArguments());
     }
+    /**
+     * Check the provided value against any range constraints specified in the Argument annotation
+     * for the corresponding field. Throw an exception if limits are violated.
+     *
+     * - Only checks numeric types (int, double, etc.)
+     */
+    private void checkArgumentRange(final ArgumentDefinition argumentDefinition, final Object argumentValue) {
+        // Only validate numeric types
```
cmnbroad: Please add a comment here saying that we only check numeric types because we have already ensured that only numeric types have bounds.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
```diff
@@ -749,6 +749,82 @@ private static String captureSystemStream(Runnable runnable, PrintStream stream,
...
+    )
+    public class WithBoundariesArguments {
+
+        @Argument(doc = "Double argument in the range [10, 20]", optional = true, minValue = 10, maxValue = 20)
+        public Double doubleArg = 15d;
+
+        @Argument(doc = "Integer in the range [0, 30]", optional = true, minValue = 0, maxValue = 30)
+        public int integerArg = 20;
+
+        @Argument(doc = "String argument", optional = true)
```
cmnbroad: This field seems redundant and out of place now with the most recent changes ?

src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
magicDGS: Done before the construction of `OptionDefinition` in `handleOptionAnnotation()`. Removed TODO here.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
```diff
@@ -749,6 +749,82 @@ private static String captureSystemStream(Runnable runnable, PrintStream stream,
...
+    public void testWithBoundariesArgumentsForNoNumeric() {
+        @CommandLineProgramProperties(summary = "broken tool",
+                oneLineSummary = "broken tool",
+                programGroup = TestProgramGroup.class)
+        class WithBoundariesArgumentsForString {
+            @Argument(doc = "String argument", optional = true, minValue = 0, maxValue = 30)
+            public String stringArg = "string";
+        }
+        final CommandLineArgumentParser clp = new CommandLineArgumentParser(new WithBoundariesArgumentsForString());
+        clp.parseArguments(System.err, new String[]{});
```
magicDGS: Done

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
magicDGS: Added recommended values to the current test.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -998,9 +1056,31 @@ public ArgumentDefinition(
...
+            this.minRecommendedValue = annotation.minRecommendedValue();
+            // bounds should be only set for numeric arguments and if the type is integer it should
+            // be set to an integer
+            this.type = CommandLineParser.getUnderlyingType(this.field);
+            if (! Number.class.isAssignableFrom(this.type)) {
+                if (this.maxValue != Double.POSITIVE_INFINITY
+                        || this.minValue != Double.NEGATIVE_INFINITY
+                        || this.maxRecommendedValue != Double.POSITIVE_INFINITY
+                        || this.minRecommendedValue != Double.NEGATIVE_INFINITY) {
+                    throw new CommandLineException.CommandLineParserInternalException("Bounds for arguments could only be set for numeric arguments");
```
magicDGS: Done

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
magicDGS: Done (both)

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
magicDGS: I maintain it for checking if it does not blows up if the argument is a `String`, but it's true that it is redundant with other tests already implemented. Removed!

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -508,6 +513,52 @@ private void validatePluginArguments() {
...
+        // Only validate numeric types because we have already ensured at constructor time that only numeric types have bounds
+        if (!Number.class.isAssignableFrom(argumentDefinition.type)) {
+            return;
+        }
+
+        final Double argumentDoubleValue = (argumentValue == null) ? null : ((Number)argumentValue).doubleValue();
+
+        // Check hard limits first, if specified
+        if (isOutOfRange(argumentDefinition.minValue, argumentDefinition.maxValue, argumentDoubleValue)) {
+            throw new CommandLineException.OutOfRangeArgumentValue(argumentDefinition.fullName, argumentDefinition.minValue, argumentDefinition.maxValue, argumentValue);
```
magicDGS: I realized that I'm using here the `argumentDefinition.fullName` field, but I guess that it is better to use the `getLongName()`, no?

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -508,6 +513,52 @@ private void validatePluginArguments() {
...
+        }
+        // Check recommended values
+        if (isOutOfRange(argumentDefinition.minRecommendedValue, argumentDefinition.maxRecommendedValue, argumentDoubleValue)) {
+            final boolean outMinValue = argumentDefinition.minRecommendedValue != Double.NEGATIVE_INFINITY;
+            final boolean outMaxValue = argumentDefinition.maxRecommendedValue != Double.POSITIVE_INFINITY;
+            if (outMinValue && outMaxValue) {
+                logger.warn("Argument --{} has value {}, but recommended within range ({},{})",
+                        argumentDefinition.fullName, argumentDoubleValue, argumentDefinition.minRecommendedValue, argumentDefinition.maxRecommendedValue);
```
magicDGS: The same here and other warnings.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -1068,15 +1069,15 @@ public ArgumentDefinition(
...
+                    throw new CommandLineException.CommandLineParserInternalException(String.format("Min/max value ranges can only be set for numeric arguments. Argument --%s has a minimum or maximum value but has a non-numeric type.", this.getLongName()));
                 }
             }
             if (Integer.class.isAssignableFrom(this.type)) {
                 if (!isInfinityOrMathematicalInteger(this.maxValue)
                         || !isInfinityOrMathematicalInteger(this.minValue)
                         || !isInfinityOrMathematicalInteger(this.maxRecommendedValue)
                         || !isInfinityOrMathematicalInteger(this.minRecommendedValue)) {
-                    throw new CommandLineException.CommandLineParserInternalException("Bounds for integer arguments should be mathematical integers");
+                    throw new CommandLineException.CommandLineParserInternalException(String.format("Integer argument --%s has a minimum or maximum value but has a non-integer type.", this.getLongName()));
```
cmnbroad: How  about "has a minimum or maximum value but has a non-integer type" -> "has a minimum or maximum attribute with a non-integral value"

src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
```diff
@@ -807,6 +809,19 @@ private void handleOptionAnnotation(final Field field, final int fieldPosition)
                 printOrder = fieldPosition * 1000;
             }
 
+            // log a warning if boundaries are set
+            if (optionAnnotation.maxValue() != Double.POSITIVE_INFINITY) {
+                logger.warn("UNSUPPORTED: maximum allowed value for argument --{} is not forced", field.getName());
```
cmnbroad: Thanks for adding these. I would remove the phrase "UNSUPPORTED: "  from all four of these, since we really are just issuing a warning. Also, I think these messages should use the argument name, not the field name (sorry if I misled you in the last round of comments). Finally, "forced" should -> "enforced" in these strings.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
```diff
@@ -769,39 +780,34 @@ public void testWithBoundariesArgumentsForNoNumeric() {
     )
     public class WithBoundariesArguments {
 
-        @Argument(doc = "Double argument in the range [10, 20]", optional = true, minValue = 10, maxValue = 20)
+        @Argument(doc = "Double argument in the range [10, 20]", optional = true, minValue = 10, minRecommendedValue = 16, maxRecommendedValue = 17, maxValue = 20)
```
cmnbroad: Can you add a comment to these saying that the recommended values here are not explicitly verified by the tests, but do force the code through the warning code paths.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
cmnbroad: Yes, that would be better.

magicDGS: Done

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java

src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
magicDGS: I removed the phrase "UNSUPPORTED" and use "enforced", but I think that the argument name is the field name for the legacy parser. Am I wrong?

magicDGS: I moved this warnings down and use `OptionDefinition.name` to be clearer here. I guess that's what you mean.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
magicDGS: Done.

magicDGS:
I addressed the latest comments. Back to you @cmnbroad and many thanks for reviewing!

src/main/java/org/broadinstitute/barclay/argparser/CommandLineException.java
```diff
@@ -39,6 +41,36 @@ public BadArgumentValue(String message) {
...
+            final DoubleFunction<String> toString = (asInt) ? Double::toString : v -> Integer.toString((int) Math.rint(v));
+            if (outMinValue && outMaxValue) {
+                return String.format("allowed range [%s, %s].", toString.apply(minValue), toString.apply(maxValue));
+            } else if (outMinValue) {
+                return String.format("minimum allowed value %s", toString.apply(minValue));
+            } else if (outMaxValue) {
+                return String.format("maximum allowed value %s", toString.apply(maxValue));
+            }
+            // this should never be reached
+            return "unknown";
```
cmnbroad: @magicDGS I know you asked about this in the last round but I never commented. I think we should throw here since it shouldn't ever happen.

magicDGS: Done

magicDGS:
Sorry for the delay, I was on holidays. I squashed/rebased my commits and included the exception thrown. Back to you,  @cmnbroad, and thanks for reviewing.

magicDGS:
Thanks for this, @cmnbroad!

PR #14, opened 2016-12-23
Improved method for finding tools.jar dependency.

cmnbroad:
Inspired by the Picard build.


droazen:
:+1: merge when ready

cmnbroad:
On my dev machine, the ToolProvider call in the gradle file returns a list of about 10 jar files, one of which is tools.jar.

PR #19, opened 2017-01-13
Handle @Hidden/@Advance annotations in the CL help

### LINKED ISSUE #6: @Hidden, @Advanced annotations are not handled by command line parser

magicDGS:
Fixes partially #6.

## EDITED:
Fixes #6 completely: 
- New special option for show `@Hidden` arguments in the CL help
- Remove @Hidden arguments if it is not specified
- Print a new argument block for `@Advanced` arguments


magicDGS:
Very simple patch to handle the @Hidden annotation in CLP. Can you review @cmnbroad?

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -291,7 +291,7 @@ public void usage(final PrintStream stream, final boolean printCommon) {
             if (optArgs != null && !optArgs.isEmpty()) {
                 stream.println("\nOptional Arguments:\n");
                 optArgs.sort(ArgumentDefinition.sortByLongName);
-                optArgs.stream().forEach(argumentDefinition -> printArgumentUsage(stream, argumentDefinition));
+                optArgs.stream().filter(argDef -> !argDef.isHidden).forEach(argumentDefinition -> printArgumentUsage(stream, argumentDefinition));
```
cmnbroad: This should have a test.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -1080,6 +1081,11 @@ public ArgumentDefinition(
                     throw new CommandLineException.CommandLineParserInternalException(String.format("Integer argument --%s has a minimum or maximum attribute with a non-integral value.", this.getLongName()));
                 }
             }
+            this.isHidden = field.getAnnotation(Hidden.class) != null;
+            if (this.isHidden && !this.optional) {
+                // required arguments cannot be optional, because they should be provided in the command line
+                throw new CommandLineException.CommandLineParserInternalException(String.format("Required argument --%s could not be hidden for the command line", this.getLongName()));
```
cmnbroad: Good catch. Can you change this string to say something like "A required argument cannot be annotated with @Hidden: %s".

cmnbroad: Also we should have a test for this.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
cmnbroad: Also, I think the filter should be moved up to the beginning of this method, where it will filter all arguments right from the start.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
magicDGS: Done both.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java

magicDGS:
Addressed comments. Back to you @cmnbroad.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
```diff
@@ -831,6 +831,41 @@ public void testOutOfRangesArguments(final String[] argv) throws Exception {
         clp.parseArguments(System.err, argv);
     }
 
+    @Test(expectedExceptions = CommandLineException.CommandLineParserInternalException.class)
+    public void testHiddenRequiredArgumentThrowException() throws Exception {
+        @CommandLineProgramProperties(summary = "tool with required and hidden argument",
+                oneLineSummary = "tool with required and hidden argument",
+                programGroup = TestProgramGroup.class)
+        final class ToolWithRequiredHiddenArgument {
+            // TODO: this does not blow up if it is a primitive, because it is not consider required
```
magicDGS: Is this a bug, @cmnbroad? If an argument is a primitive it is considered optional independently of the optional flag...

cmnbroad: Its a known issue, but results in somewhat surprising behavior. The parser treats any argument that has a non-null value as if it were "provided". Since primitives are never null, the parser is always satisfied that they've been set, which makes them appear to be optional even when they're required. The comment for the optional attribute in Argument describes this behavior.

lbergelson:
Could we add an option that displays hidden arguments?  I've run into issues where I've had to search gatk3 source code because an argument exists but it is hidden and won't display in the help.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
magicDGS: Ok, I overseen that primitives are never null and this is point out in the javadoc. I think that this could be fixed (in a different PR) by checking if the field has a primitive type to determine if it is optional or not. I removed this comment and I will open an issue.

magicDGS:
I rebased to include the changes in #22 and I included a param in usage for print the Hidden arguments. Can you have a look @lbergelson to see if it is what you want?

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -387,7 +389,8 @@ public boolean parseArguments(final PrintStream messageStream, final String[] ar
 
         //check if special short circuiting arguments are set
         if (isSpecialFlagSet(parsedArguments, SpecialArgumentsCollection.HELP_FULLNAME)) {
-            messageStream.print(usage(true));
+            // TODO: should a new SpecialArgumentCollection being set to display hidden arguments?
```
magicDGS: Should I include this here, @lbergelson?

src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
```diff
@@ -1050,6 +1055,9 @@ private OptionDefinition(final Field field, final String name, final String shor
             this.isCommon = isCommon;
             this.mutuallyExclusive = new HashSet<String>(Arrays.asList(mutuallyExclusive));
             this.printOrder = printOrder;
+            if (this.field.getAnnotation(Hidden.class) != null) {
+                logger.warn("Hidden argument --{} is not handled", this.name);
```
magicDGS: Should `@Hidden` be handled in legacy, @cmnbroad? It's a trivial change.

magicDGS:
Friendly ping @lbergelson and @cmnbroad

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
cmnbroad: @magicDGS Yes, I think we want to add special argument called "showHidden", and annotate it as....Advanced. We'd then need to also add support for Advanced args in the usage output, so they show up in a special group (we need to do that anyway, but this will force our hand - if this is more than you signed up for I'll be happy to do it). @vdauwera any input on what the language should be for the "Advanced" argument group should be, maybe "For advanced/experimental use only" or something ?

src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
cmnbroad: I don't think Hidden or Advanced were present or used in Picard, and we don't really want to invest in adding more features to the old parser, so I think what you've done here is exactly the right thing.

magicDGS: Ok.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
magicDGS: I added the "showHidden" special argument, and it is annotated with Advanced. I'm up for handling `@Advance` in the command line (actually it was in my plans), but maybe it will be better to do it in a different PR to get this into.

magicDGS:
Back to you @cmnbroad.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
```diff
@@ -831,6 +831,42 @@ public void testOutOfRangesArguments(final String[] argv) throws Exception {
...
+        new CommandLineArgumentParser(new ToolWithRequiredHiddenArgument());
+    }
+
+    @Test
+    public void testHiddenArguments() throws Exception {
+        @CommandLineProgramProperties(summary = "test",
+                oneLineSummary = "test",
+                programGroup = TestProgramGroup.class)
+        final class ToolWithHiddenArgument {
+            @Argument(fullName = "hidden", shortName = "hidden", doc = "Hidden argument", optional = true)
```
cmnbroad: Can you change the fullName of this argument to something more unique and less likely to show up in the usage output by accident, like "hiddenTestArgument".

magicDGS: Done.

magicDGS:
I included 4 more commits for:

1. Address your comment regarding the arg name in the test
2. Add a helper method for print an usage block (may be useful in the future)
3. Handle @Advance annotations by printing them in the usage after optional arguments
4. Update a couple of tests to check for this new block

Back to you @cmnbroad. Let me know if this looks like the `@Advance` handling that you were referring to.

magicDGS:
Any advance on this, @cmnbroad? Thank you in advance!

cmnbroad:
@magicDGS Also can you squash all of these down into one commit, and then add a single additional commit with your changes for this review round. It makes reviewing much easier. Thx.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
```diff
@@ -132,52 +133,58 @@ public void testAbbreviationsAreRejected() {
     public void testOptionalOnlyUsage() {
         final OptionalOnlyArguments oo = new OptionalOnlyArguments();
         final CommandLineArgumentParser clp = new CommandLineArgumentParser(oo);
-        final String out = clp.usage(false); // without common args
+        final String out = clp.usage(false, false); // without common/hidden args
         final int reqIndex = out.indexOf("Required Arguments:");
         Assert.assertTrue(reqIndex < 0);
         Assert.assertTrue(out.indexOf("Optional Arguments:", reqIndex) > 0);
         Assert.assertEquals(out.indexOf("Conditional Arguments:", reqIndex), -1);
+        Assert.assertEquals(out.indexOf("Advanced Arguments:", reqIndex), -1);
```
magicDGS: @cmnbroad, this is a test for non-advance arguments not printing "Advanced Arguments" string prefix if there are none.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
```diff
@@ -132,52 +133,58 @@ public void testAbbreviationsAreRejected() {
...
+        validateRequiredOptionalUsage(clp, false, true); // without common args
     }
 
     @Test
     public void testWithoutPositionalWithDefaultUsage() {
         final ArgumentsWithoutPositional fo = new ArgumentsWithoutPositional();
         final CommandLineArgumentParser clp = new CommandLineArgumentParser(fo);
-        validateRequiredOptionalUsage(clp, true); // with common args
+        // does not have the special showHidden advanced argument
+        validateRequiredOptionalUsage(clp, true, false); // with common args
```
magicDGS: This is also a test for not printing "Advanced Arguments" prefix.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
```diff
@@ -132,52 +133,58 @@ public void testAbbreviationsAreRejected() {
...
     }
 
     @Test
     public void testWithoutPositionalWithoutDefaultUsage() {
         final ArgumentsWithoutPositional fo = new ArgumentsWithoutPositional();
         final CommandLineArgumentParser clp = new CommandLineArgumentParser(fo);
-        validateRequiredOptionalUsage(clp, false); // without commoon args
+        // does not have the special showHidden advanced argument
+        validateRequiredOptionalUsage(clp, false, false); // without common args
```
magicDGS: This is also a test for not printing "Advanced Arguments" prefix.

magicDGS:
Squashed and rebased to the latest master. I implemented the test case for the advanced/required combination, and point out in comments the tests for not printing the prefix for advanced arguments. Back to you @cmnbroad!

src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
cmnbroad: Its the annotation thats not handled, not the argument. This should probably say something like "Hidden annotation is not honored for --{}".

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
```diff
@@ -132,52 +133,58 @@ public void testAbbreviationsAreRejected() {
...
+        Assert.assertEquals(out.indexOf("Advanced Arguments:", reqIndex) != -1, hasAdvanced);
     }
 
     @Test
     public void testRequiredOptionalWithDefaultUsage() {
         final FrobnicateArguments fo = new FrobnicateArguments();
         final CommandLineArgumentParser clp = new CommandLineArgumentParser(fo);
-        validateRequiredOptionalUsage(clp, true); // with common args
+        // have the special showHidden advanced argument
+        validateRequiredOptionalUsage(clp, true, true); // with common args
```
magicDGS: This is validating if the "Advanced Arguments:" prefix is printed, @cmnbroad. Because one of the special arguments is advanced (`--showHidden`), the `FrobnicateArguments` should print this out.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
```diff
@@ -132,52 +133,58 @@ public void testAbbreviationsAreRejected() {
...
     }
 
     @Test
     public void testRequiredOptionalWithoutDefaultUsage() {
         final FrobnicateArguments fo = new FrobnicateArguments();
         final CommandLineArgumentParser clp = new CommandLineArgumentParser(fo);
-        validateRequiredOptionalUsage(clp, false); // without common args
+        // have the special showHidden advanced argument
+        validateRequiredOptionalUsage(clp, false, true); // without common args
```
magicDGS: The same here for printing "Advance Arguments" prefix.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
cmnbroad: I see - I couldn't understand the comment before, and since the dependency is not explicit in this file, its hard to see whats going on.  I would ask you to do two things to make this all a bit clearer:

1) Change the comments that say "have the special showHidden advanced argument" to say "FrobnicateArguments has a SpecialArgumentsCollection that contains an @Advanced argument ("called showHidden")). That will make whats going on with these test much more transparent.

2) As far as I can tell, we still have no positive test to ensure that the actual advanced arguments are displayed. Can you please add a separate test just like the one you added for Hidden, where you test not only the prefix, but that the actual argument is being displayed.

magicDGS: Done.

magicDGS:
I think that everything is addressed now. Thanks a lot for reviewing!

cmnbroad:
@magicDGS Thanks!

magicDGS:
Thank to you @cmnbroad!

PR #21, opened 2017-01-16
Separate optional tool-specific args from common args in usage output.

cmnbroad:
This groups all the tool-specific args together by separating the optional **_tool_** args from the optional **_common_** args by changing the command line usage output groupings from:

- Required args
- **_All_** Optional args
- Plugin args

to:

- Required args
- Optional **_tool_** args
- Optional **_common_** args
- Plugin args

Docgen already does this for generated docs.


lbergelson:
@cmnbroad What did we decide to do here?  I think the consensus was to ask Geraldine.

cmnbroad:
Yes, we shouldn't review or merge this for now.

PR #22, opened 2017-01-16
Change CommandLineParser interface usage method signature to return a string.

cmnbroad:
Note that although this will affect Picard as well, it will happen at the same time as the other Barclay changes, so it should be ok. I didn't spend any time beefing up the legacy parser usage tests, which are pretty much just smoke tests; I just update them to reflect the new method signature.


lbergelson:
👍

PR #24, opened 2017-01-18
Change internal command line exception classes to package-protected RuntimeExceptions

cmnbroad:
and change their base class to RuntimeException so that all CommandLineExceptions can be treated as user errors.

### DIFF
--- a/src/main/java/org/broadinstitute/barclay/argparser/CommandLineException.java
+++ b/src/main/java/org/broadinstitute/barclay/argparser/CommandLineException.java
@@ -76,8 +76,9 @@ public class CommandLineException extends RuntimeException {
      * Class CommandLineParserInternalException
      * <p/>
      * For internal errors in the command line parser not related to syntax errors in the command line itself.
+     * Package protected to restrict usage to the arg parsers.
      */
-    public static class CommandLineParserInternalException extends CommandLineException {
+    static class CommandLineParserInternalException extends RuntimeException {
         private static final long serialVersionUID = 0L;
         public CommandLineParserInternalException( final String s ) {
             super(s);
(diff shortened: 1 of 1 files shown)


src/main/java/org/broadinstitute/barclay/argparser/CommandLineException.java
```diff
@@ -76,8 +76,9 @@ private static String getMessage(final double minValue, final double maxValue, f
      * Class CommandLineParserInternalException
      * <p/>
      * For internal errors in the command line parser not related to syntax errors in the command line itself.
+     * Package protected to restrict usage to the arg parsers.
      */
-    public static class CommandLineParserInternalException extends CommandLineException {
+    static class CommandLineParserInternalException extends RuntimeException {
```
droazen: Should probably be public to allow clients to catch it.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineException.java
```diff
@@ -90,16 +91,17 @@ public CommandLineParserInternalException( final String s, final Throwable throw
 
     /**
      * For wrapping errors that are believed to never be reachable
+     * Package protected to restrict usage to the arg parsers.
      */
-    public static class ShouldNeverReachHereException extends CommandLineException {
+    static class ShouldNeverReachHereException extends RuntimeException {
```
droazen: Should probably extend `CommandLineParserInternalException`, which should be the superclass of all internal errors.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineException.java
cmnbroad: Done.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineException.java
cmnbroad: Done.

droazen:
:+1: merge when ready

Issue #25, opened 2017-01-24
Implement the generic ability for Collection arguments to be provided via a file


droazen:
For any argument of a Collection type, we want the ability to provide a `.list` file containing the literal values for the argument, instead of having to provide all literal values on the command line. 

So, if an argument is of a `Collection` type, and the token from the command line ends in `.list`, treat the token as a file name, load it, and unpack all lines in the file into the `Collection`.

droazen:
For @cmnbroad

PR #28, opened 2017-01-26
Add support for collection list files on the command line.

cmnbroad:
Fixes https://github.com/broadinstitute/barclay/issues/25.

### DIFF
--- a/src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
+++ b/src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
@@ -69,6 +69,9 @@ public final class CommandLineArgumentParser implements CommandLineParser {
     public static final String COMMENT = "#";
     public static final String POSITIONAL_ARGUMENTS_NAME = "Positional Argument";
 
+    // Extension for collection argument list files
+    private static final String COLLECTION_LIST_FILE_EXTENSION = ".list";
+
     private static final Logger logger = LogManager.getLogger();
 
     // Map from (full class) name of each CommandLinePluginDescriptor requested and
@@ -581,7 +584,7 @@ public final class CommandLineArgumentParser implements CommandLineParser {
     }
 
     @SuppressWarnings("unchecked")
-    private void setArgument(ArgumentDefinition argumentDefinition, final List<String> values) {
+    private void setArgument(ArgumentDefinition argumentDefinition, List<String> values) {
         //special treatment for flags
         if (argumentDefinition.isFlag() && values.isEmpty()){
             argumentDefinition.hasBeenSet = true;
@@ -598,6 +601,7 @@ public final class CommandLineArgumentParser implements CommandLineParser {
             @SuppressWarnings("rawtypes")
             final Collection c = (Collection) argumentDefinition.getFieldValue();
             c.clear();
+            values = expandListFile(values);
         }
 
         for (String stringValue: values) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -69,6 +69,9 @@
     public static final String COMMENT = "#";
     public static final String POSITIONAL_ARGUMENTS_NAME = "Positional Argument";
 
+    // Extension for collection argument list files
+    private static final String COLLECTION_LIST_FILE_EXTENSION = ".list";
```
magicDGS: Maybe this should be added to the usage line for collections...

lbergelson: I agree that this functionality needs to be discoverable somehow... Adding it to every Collection usage seems redundant, but maybe maybe not?  The help output definitely has to include is as an option somewhere, either in the preamble or on every collection argument.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -639,6 +643,49 @@ private void setArgument(ArgumentDefinition argumentDefinition, final List<Strin
...
+    /**
+     * Read a list file and return a list of the collection values contained in it
+     * A line that starts with {@link #COMMENT}  is ignored.
+     *
+     * @param collectionListFile a text file containing list values
+     * @return false if a fatal error occurred
+     */
+    private List<String> loadCollectionListFile(final String collectionListFile) {
+        List<String> args = new ArrayList<>();
+        try (BufferedReader reader = new BufferedReader(new FileReader(collectionListFile))){
```
lbergelson: This is one of those places that I think the stream api really improves things:

``` java
        try (BufferedReader reader = new BufferedReader(new FileReader(collectionListFile))){
            return reader.lines()
                    .map(String::trim)
                    .filter(line -> !line.isEmpty())
                    .filter(line -> !line.startsWith(COMMENT))
                    .collect(Collectors.toList());
        } catch (final IOException e) {
            throw new CommandLineException("I/O error loading list file:" + collectionListFile, e);
        }
```

cmnbroad: I like it.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
cmnbroad: Also agree its not discoverable, and that we should do something, but the same is true for using "null" to reset collections, regular arguments files, comments, etc.. The usage output is already intimidating, and thats just for PrintReads! I wonder if we should maybe add a link to an online help document about using the CLP ? @vdauwera any opinions on that ?

lbergelson: Maybe we need a `--helphelp` option.  In all seriousness though, being able to control help verbosity would be nice... We could do the` -h`, `-hh`, `-hhhhhhhh` thing that some people use for verbosity, although that's a bit funky.

cmnbroad: We may need something like that. Especially since meanwhile, on another branch, I'm adding tagged arguments...

magicDGS: This will be maybe better to be ".arg_list" to avoid possible classes with a input file that represent a list of elements (like intervals).

cmnbroad:
I created a separate [ticket](https://github.com/broadinstitute/barclay/issues/35) to cover integration of this feature (and some others) wth docgen/usage output.

lbergelson:
@magicDGS `.list` has been the standard in gatk3 for a while.  Are there any specific .list extension files that you know about and work with?

magicDGS:
Not really, but It is a very generic extension. I imagine situations where the user may specify a file with that extension for a File collection that does not require any specific extension.

Anyway, if there is a documentation at some point it may be useful. At least a check for the extension should be available (made public the static field).

cmnbroad:
@lbergelson I've squashed and rebased. Anything else before I merge ?

lbergelson:
@cmnbroad 👍

PR #32, opened 2017-01-31
Bound getDefaultInstances

magicDGS:
As discussed in https://github.com/broadinstitute/gatk/issues/2362

### DIFF
--- a/src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
+++ b/src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
@@ -65,7 +65,7 @@ public abstract class CommandLinePluginDescriptor<T> {
      * this class in any of the packages returned by {@link #getPackageNames} will be command line
      * accessible.
      */
-    public abstract Class<?> getPluginClass();
+    public abstract Class<? super T> getPluginClass();
 
     /**
      * List of package names from which to load command line plugin classes.
@@ -80,6 +80,7 @@ public abstract class CommandLinePluginDescriptor<T> {
      * dynamically discoverable.
      * @return false if the class shouldn't be used; otherwise true
      */
+    // TODO: I don't think that this should be bounded, but I think that the default value should return c -> c instanceof getPluginClass()
     public Predicate<Class<?>> getClassFilter() { return c -> true;}
 
     /**
@@ -102,7 +103,7 @@ public abstract class CommandLinePluginDescriptor<T> {
      * @throws IllegalAccessException
      * @throws InstantiationException
      */
-    public abstract Object getInstance(Class<?> pluggableClass)
+    public abstract T getInstance(Class<?> pluggableClass)
             throws IllegalAccessException, InstantiationException;
 
     /**
@@ -135,6 +136,7 @@ public abstract class CommandLinePluginDescriptor<T> {
      * @return true if the plugin for this class was specified on the command line, or the
      * values in this class may be set byt he user, otherwise false
      */
+    // TODO: I don't think that this should be bounded
     public abstract boolean isDependentArgumentAllowed(Class<?> dependentClass);
 
     /**
@@ -173,6 +175,7 @@ public abstract class CommandLinePluginDescriptor<T> {
      * @param pluginName Name of the plugin requested
      * @return Class object for the plugin instance requested
      */
+    // TODO: this method is unused in Barclay and GATK -> it is something needed?
     public abstract Class<?> getClassForInstance(final String pluginName);
 
 }
--- a/src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
+++ b/src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
@@ -78,10 +78,12 @@ public abstract class CommandLinePluginDescriptor<T> {
     /**
      * Give this descriptor a chance to filter out any classes it doesn't want to be
      * dynamically discoverable.
+     *
+     * Default implementation uses {@link Class#isAssignableFrom(Class)}.
+     *
      * @return false if the class shouldn't be used; otherwise true
      */
-    // TODO: I don't think that this should be bounded, but I think that the default value should return c -> c instanceof getPluginClass()
-    public Predicate<Class<?>> getClassFilter() { return c -> true;}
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -161,7 +161,7 @@ public abstract Object getInstance(Class<?> pluggableClass)
      * @return the default plugins used for this instance of this descriptor as a list of Object. Used for
      * help/doc generation.
      */
-    public abstract List<Object> getDefaultInstances();
+    public abstract List<? extends T> getDefaultInstances();
```
droazen: @cmnbroad & I both think this should be `? super T`, since there's client code that will want this typed as `List<Object>`, and there's also client code that could want to type it as `List<T>`, but it seems unlikely that there would be client code that would type it as a List of a subclass of T.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -161,7 +161,7 @@ public abstract Object getInstance(Class<?> pluggableClass)
...
 
     /**
      * @return an ordered List of actual plugin instances that have been specified on the command
```
droazen: Return value of `getAllInstances()` should be consistent with that of `getDefaultInstances()`

cmnbroad: We should audit each place where a wildcard or Object appears, including the Class<> references, and tighten the bounds if appropriate. But before we merge anything I'd want to verify it by building GATK with the changes to make sure the effects on the GATK plugin descriptor are reasonable, and to get confidence that we haven't been overly restrictive.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -135,6 +136,7 @@ public abstract Object getInstance(Class<?> pluggableClass)
      * @return true if the plugin for this class was specified on the command line, or the
      * values in this class may be set byt he user, otherwise false
      */
+    // TODO: I don't think that this should be bounded
     public abstract boolean isDependentArgumentAllowed(Class<?> dependentClass);
```
magicDGS: I believe that this should be implementation-dependent.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -80,6 +80,7 @@
      * dynamically discoverable.
      * @return false if the class shouldn't be used; otherwise true
      */
+    // TODO: I don't think that this should be bounded, but I think that the default value should return c -> c instanceof getPluginClass()
     public Predicate<Class<?>> getClassFilter() { return c -> true;}
```
magicDGS: If this is a public method that some client may use, I think that it should not be bounded but at least being:
```java
public Predicate<Class<?>> getClassFilter() {
    c -> c.isAssignableFrom(getPluginClass())
}
```
Or something similar.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -161,18 +163,19 @@ public abstract Object getInstance(Class<?> pluggableClass)
...
      */
-    public abstract List<T> getAllInstances();
+    public abstract List<? super T> getAllInstances();
 
     /**
      * @param pluginName Name of the plugin requested
      * @return Class object for the plugin instance requested
      */
+    // TODO: this method is unused in Barclay and GATK -> it is something needed?
     public abstract Class<?> getClassForInstance(final String pluginName);
```
magicDGS: What is the purpose of this method?

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
cmnbroad: I'm not sure what you mean by "implementation dependent" in this case. But I think the type bounds should stay like it is since this is used by the command line parser.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
cmnbroad: I used it in an interim version of docgen, I believe because I needed access to the full classname of the plugIn in order to generate an html link. So, I think its another case where it should remain as is.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
cmnbroad: Right - I think this should also not be bounded, and I like your proposed default implementation.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
magicDGS: That's exactly what I meant.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
magicDGS: Removed TODO and changed default implementation.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
magicDGS: Removed TODO.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
magicDGS: Removed TODO.

magicDGS:
I removed the TODOs and add the new default implementation of the class filter. Back to you @cmnbroad for the decision about the rest.

cmnbroad:
Closing in favor of https://github.com/broadinstitute/barclay/pull/38. Lets keep the branch for now since there was some discussion in here. In short, I think it makes more sense to use unbounded wildcards for the Class type bounds, since not all of them can be narrowed, and bounding only some of them seems inconsistent.

Issue #29, opened 2017-01-26
Add support for argument tagging to barclay


droazen:
Currently in GATK4 we have some support for tagging arguments, but the tagging is done on the value side, and manually by the engine. Eg.,

```
-V myTag:my.vcf
```

This causes many issues: parsing ambiguity with URIs, interfering with shell auto-expansion, etc.

Let's add native tagging support to barclay, so that the tag can be on the argument side rather than the value side. Eg.,

```
-I:tumor my.bam
```

We also need to support arbitrary key/value pairs after the tag. Eg., 

```
-I:tumor,key=value,key2=value2
```

I think that the way to do this is to introduce a new `TaggedArgument` interface in barclay with methods to get/set both the tag and the optional key/value pairs. Eg.,

```
void setTag(String)
String getTag()
void setTagAttributes(Map<String, String> attributes)
Map<String, String> getTagAttributes()
```

The upcoming `URI` class in GATK will implement this interface. When barclay encounters an `@Argument`-annotated field of a type that implements the `TaggedArgument` interface, it should parse the tags and inject them into the object instance for that field via the setter methods.

droazen:
For @cmnbroad

cmnbroad:
Should we bake in a way to support @LeeTL1220's time series example ? I'm not sure how many there will be in a series; if its a only a few, the tool might be able to use predefined names. Otherwise we may want some order-preserving mechanism.

cmnbroad:
Also, we might want to consider adding a way to declare the allowed tags and/or attributes, maybe in Argument or a separate annotation, so instead of the CLP displaying:

     -I,-I:FeatureInput

it could display something like:

      -I,-I: File
      Allowed tags: tumor, normal
      Allowed attributes truth, training

droazen:
These are good ideas for future enhancements, but let's get the basic version described above in first -- we'll need it for hookup to GATK by early next week.

cmnbroad:
@droazen Some questions on how this should work (from the CLP's perspective), and how to handle command line validation:

- Are attributes are always optional ?
- Is tag name always required for an option that implements `TaggedArgument` (I'm not sure FeatureInput [handles this](https://github.com/broadinstitute/gatk/issues/2144) correctly) ?
- Should tags be allowed for non-collections ?
- Should the CLP require tag names to be unique within a collection, or should we leave that up to the tool ?

lbergelson:
@cmnbroad What's the time series use case?  

[...]

PR #33, opened 2017-01-31
Add tagged argument support.

cmnbroad:
Fixes https://github.com/broadinstitute/barclay/issues/29.

### DIFF
--- a/src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
+++ b/src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
@@ -8,6 +8,7 @@ import joptsimple.OptionSpecBuilder;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.text.WordUtils;
 
+import org.apache.commons.lang3.tuple.Pair;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.broadinstitute.barclay.utils.Utils;
@@ -29,7 +30,6 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.function.DoubleFunction;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -77,6 +77,9 @@ public final class CommandLineArgumentParser implements CommandLineParser {
     // Map from (full class) name of each CommandLinePluginDescriptor requested and
     // found to the actual descriptor instance
     private Map<String, CommandLinePluginDescriptor<?>> pluginDescriptors = new HashMap<>();
+
+    // Keeps a map of tagged arguments for just-in-time retrieval at field population time
+    private TaggedArgumentParser tagParser = new TaggedArgumentParser();
 
     // Return the plugin instance corresponding to the targetDescriptor class
     @Override
@@ -129,10 +132,6 @@ public final class CommandLineArgumentParser implements CommandLineParser {
     // Maps long name, and short name, if present, to an argument definition that is
     // also in the argumentDefinitions list.
     private final Map<String, ArgumentDefinition> argumentMap = new LinkedHashMap<>();
-
-    // In case implementation wants to get at arg for some reason.
-    private String[] argv;
-
 
     // The associated program properties using the CommandLineProgramProperties annotation
     private final CommandLineProgramProperties programProperties;
@@ -345,8 +344,13 @@ public final class CommandLineArgumentParser implements CommandLineParser {
      */
     @SuppressWarnings("unchecked")
     @Override
-    public boolean parseArguments(final PrintStream messageStream, final String[] args) {
-        this.argv = args;
+    public boolean parseArguments(final PrintStream messageStream, String[] args) {
+
+        // Preprocess the arguments before the parser sees them, replacing any tagged options
+        // and their values with raw option names and surrogate key values, so that tagged
+        // options can be recognized by the parser. The actual values will be retrieved using
+        // the key when the fields's values are set.
+        args = tagParser.preprocessTaggedOptions(args);
 
         OptionParser parser = new OptionParser(false);
 
@@ -619,7 +623,40 @@ public final class CommandLineArgumentParser implements CommandLineParser {
                     throw new CommandLineException("Non \"null\" value must be provided for '" + argumentDefinition.getNames() + "'.");
                 }
             } else {
-                value = constructFromString(CommandLineParser.getUnderlyingType(argumentDefinition.field), stringValue, argumentDefinition.getLongName());
+                // See if the value is a surrogate key in the tag parser's map that was placed there during preprocessing,
+                // and if so, unpack the values retrieved via the key and use those to populate the field
+                Pair<String, String> taggedOptionPair = tagParser.getTaggedOptionForSurrogate(stringValue);
+                if (TaggedArgument.class.isAssignableFrom(argumentDefinition.type)) {
+                    value = constructFromString(
+                            CommandLineParser.getUnderlyingType(argumentDefinition.field),
+                            taggedOptionPair == null ?
+                                    stringValue :
+                                    taggedOptionPair.getRight(),        // argument value
+                            argumentDefinition.getLongName());
+                    // NOTE: this propagates the tag name/attributes to the field BEFORE the value is set
+                    TaggedArgument taggedArgument = (TaggedArgument) value;
+                    tagParser.populateArgumentTags(
+                            taggedArgument,
+                            argumentDefinition.getLongName(),
+                            taggedOptionPair == null ?
+                                    argumentDefinition.getLongName() :
+                                    taggedOptionPair.getLeft());
+                }
+                else {
+                    if (taggedOptionPair == null) {
+                        value = constructFromString(
+                                CommandLineParser.getUnderlyingType(argumentDefinition.field),
+                                stringValue,
+                                argumentDefinition.getLongName());
+                    } else {
+                        // a tag was found for a non-taggable argument
+                        throw new CommandLineException(
+                                String.format("The argument: \"%s/%s\" does not accept tags: \"%s\"",
+                                        argumentDefinition.shortName,
+                                        argumentDefinition.fullName,
+                                        taggedOptionPair.getLeft()));
+                    }
+                }
             }
 
             // check the argument range
@@ -1193,7 +1230,12 @@ public final class CommandLineArgumentParser implements CommandLineParser {
                 if (isSensitive){
                     return String.format("--%s ***********", getLongName());
                 } else {
-                    return String.format("--%s %s", getLongName(), value);
+                    if (value instanceof TaggedArgument) {
+                        TaggedArgument taggedArg = (TaggedArgument) value;
+                        return String.format("--%s %s", TaggedArgumentParser.getDisplayString(getLongName(), taggedArg), value);
+                    } else {
+                        return String.format("--%s %s", getLongName(), value);
+                    }
                 }
             }
             return "";
--- a/src/main/java/org/broadinstitute/barclay/argparser/TaggedArgument.java
+++ b/src/main/java/org/broadinstitute/barclay/argparser/TaggedArgument.java
@@ -1,0 +1,39 @@
+package org.broadinstitute.barclay.argparser;
+
+import java.util.Map;
+
+/**
+ * Interface for arguments that can have tags and attributes. The command line argument parser
+ * looks for argument fields that implement this interface, and propagates tag names and
+ * attribute/value pairs to the field.
+ *
+ * NOTE: No check is done to prevent duplicate tag names from being used more than once
+ * in an arugment that is a Collection.
+ */
+public interface TaggedArgument {
+
+    /**
+     * Set the tag name (optional - required only if attributes are present) for this instance.
+     * @param tagName
+     */
+    void setTag(String tagName);
+
+    /**
+     * Retrieve the tag name for this instance.
+     * @return String representing the tagName. Should not return null.
+     */
+    String getTag();
+
+    /**
+     * Set the attribute/value pairs for this instance.
+     * @param attributes Map of attribute names and values for this arguments.
+     */
+    void setTagAttributes(Map<String, String> attributes);
+
+    /**
+     * Get the attribute/value pair Map for this instance. May be empty.
+     * @return Map of attribute/value pairs for this instance.
+     */
+    Map<String, String> getTagAttributes();
+
+}
--- a/src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
+++ b/src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
@@ -1,0 +1,321 @@
+package org.broadinstitute.barclay.argparser;
+
+import org.apache.commons.lang3.tuple.Pair;
+import org.broadinstitute.barclay.utils.Utils;
+
+import java.util.*;
+
+/**
+ * The parser used by the {@link CommandLineArgumentParser} for handling tagged argument strings. Tagged arguments
+ * can optionally contain a logical name, and optional attribute/value pairs. They are are of the form:
+ *
+ *     --argument_name:logical_name argument_value
+ * or:
+ *     --argument_name:logical_name,key1=value1,key2=value2 argument_value
+ *
+ * The logical name is optional, but is required if key/value pairs are includede. @Argument annotated fields that
+ * require tagged values must implement {@link TaggedArgument}.
+ *
+ * The parsing occurs in two phases:
+ *
+ * 1) Phase 1 occurs before the option parser is presented with the arguments. In phase 1, the option name is
+ * peeled off from the raw option string (which includes that tag string), and the resulting tag string and accompanying
+ * raw argument value are stored as a pair in a hash map for later retrieval via a key. The key is a string that is
+ * constructed by concatenating the original option name (excluding the actual short/long hyphen prefix), the
+ * tag string, and the raw argument value together to ensure the key is unique:
+ *
+ *      Key: --argument_name:logical_name,key1=value1,key2=value2:argument_value
+ *
+ * The pair object stored in the map contains only the (previously peeled off) tag string, and the raw argument value that
+ * was provided by the user. The command line parser replaces the original arguments provided by the user with the option
+ * name and the key:
+ *
+ *          User provides:              "--argument_name:logical_name,key1=value1,key2=value2" "raw_argument_value"
+ *          Parser is presented with:   "--argument_name" "argument_name:logical_name,key1=value1,key2=value2:argument_value"
+ *          Map stores:                 Pair("logical_name,key1=value1,key2=value2", "argument_value")
+ *
+ * 2) In phase 2, which occurs when the underlying argument field is being populated with a value, the key is used to
+ * retrieve the original tag string and argument value. The tag string is parsed (logical name and attributes) and used
+ * to populate the underlying argument field.
+ */
+public final class TaggedArgumentParser {
+
+    /**
+     * Delimiter between key-value pairs in the "logical_name,key1=value1,key2=value2" syntax.
+     */
+    private static final String ARGUMENT_KEY_VALUE_PAIR_DELIMITER = ",";
+
+    /**
+     * Separator between keys and values in the "logical_name,key1=value1,key2=value2" syntax.
+     */
+    private static final String ARGUMENT_KEY_VALUE_SEPARATOR = "=";
+
+    /**
+     * Separator used between option name and logical name.
+     */
+    private static final char ARGUMENT_TAG_NAME_SEPARATOR = ':';
+
+    private static final String USAGE = "Tagged arguments must be of the form argument_name or argument_name:logical_name(,key=value)*";
+
+    // Map of surrogate keys to Pair(tag_string, raw_argument_value)
+    private Map<String, Pair<String, String>> tagSurrogates = new HashMap<>();
+
+    // Prefixes used by the opt parser for short/long prefixes
+    private static final String SHORT_OPTION_PREFIX = "-";
+    private static final String LONG_OPTION_PREFIX = "--";
+
+    /**
+     * Given an array of raw arguments provided by the user, return an array of args where tagged arguments
+     * have been replaced with curated arguments containing a key to be used by the parser to retrieve the actual
+     * values.
+     * @param argArray raw arguments as provided by the user
+     * @return curated string of arguments to be presented to the opt parser
+     */
+    public String[] preprocessTaggedOptions(final String[] argArray) {
+        List<String> finalArgs = new ArrayList<>(argArray.length);
+
+        Iterator<String> argIt = Arrays.asList(argArray).iterator();
+        while (argIt.hasNext()) {
+            final String arg = argIt.next();
+            if (isShortOptionToken(arg)) {
+                replaceTaggedOption(SHORT_OPTION_PREFIX, arg.substring(SHORT_OPTION_PREFIX.length()), argIt, finalArgs);
+            } else if (isLongOptionToken(arg)) {
+                replaceTaggedOption(LONG_OPTION_PREFIX, arg.substring(LONG_OPTION_PREFIX.length()), argIt, finalArgs);
+            } else { // Positional arg, etc., just add it
+                finalArgs.add(arg);
+            }
+
+        }
+        return finalArgs.toArray(new String[finalArgs.size()]);
+    };
+
+    /**
+     * Process a single option and add it to the curated args list. If the option is tagged, add the
+     * curated key and value. Otherwise just add the raw option.
+     *
+     * @param optionPrefix the actual option prefix used for this option, either "-" or "--"
+     * @param optionString the option string including everything but the prefix
+     * @param userArgIt iterator of raw arguments provided by the user, used to retrieve the value corresponding
+     *                    to this option
+     * @param finalArgList the curated argument list
+     */
+    private void replaceTaggedOption(
+            final String optionPrefix,          // the option prefix used (short or long)
+            final String optionString,          // the option string, minus prefix, including any tags/attributes
+            final Iterator<String> userArgIt,   // the original, raw, un-curated input argument list
+            final List<String> finalArgList     // final, curated argument list
+    )
+    {
+        final int separatorIndex = optionString.indexOf(TaggedArgumentParser.ARGUMENT_TAG_NAME_SEPARATOR);
+        if (separatorIndex == -1) { // no tags, consume one argument and get out
+            finalArgList.add(optionPrefix + optionString);
+        } else {
+            if (userArgIt.hasNext()) {
+
+                final String optionName = optionString.substring(0, separatorIndex);
+                if (optionName.isEmpty()) {
+                    throw new CommandLineException("Zero length argument name found in tagged argument: " + optionString);
+                }
+                final String tagNameAndValues = optionString.substring(separatorIndex+1, optionString.length());
+                if (tagNameAndValues.isEmpty()) {
+                    throw new CommandLineException("Zero length tag name found in tagged argument: " + optionString);
+                }
+                final String argValue = userArgIt.next();
+                if (isLongOptionToken(argValue) || isShortOptionToken(argValue)) {
+                    // An argument value is required, and there isn't one to consume
+                    throw new CommandLineException("No value found for tagged argument: " + optionString);
+                }
+
+                // Replace the original prefix/option/attribute string with the original prefix/option name, and
+                // replace it's companion argument value with the surrogate key to be used later to retrieve the
+                // actual values
+                final String pairKey = getSurrogateKeyForTaggedOption(optionString, argValue, tagNameAndValues);
+                finalArgList.add(optionPrefix + optionName);
+                finalArgList.add(pairKey);
+            } else {
+                // the option appears to be tagged, but we're already at the end of the argument list,
+                // and there is no companion value to use
+                throw new CommandLineException("No value found for tagged argument: " + optionString);
+            }
+        }
+    }
+
+    /**
+     * Attempt to retrieve an option pair from the map using a surrogate key.
+     * @param putativeSurrogateKey putative key to try to retrieve from the surrogate map
+     * @return tagged option pair for this surrogate, or null if no entry
+     */
+    public Pair<String, String> getTaggedOptionForSurrogate(final String putativeSurrogateKey) {
+        return tagSurrogates.get(putativeSurrogateKey);
+    }
+
+    // See if the opt parser would think this is a short option ("-")
+    private static boolean isShortOptionToken(final String argument) {
+        return argument.startsWith( SHORT_OPTION_PREFIX )
+                && !SHORT_OPTION_PREFIX.equals( argument )
+                && !isLongOptionToken( argument );
+    }
+
+    // See if the opt parser would think this is a long option ("--")
+    private static boolean isLongOptionToken(final String argument) {
+        return argument.startsWith( LONG_OPTION_PREFIX );
+    }
+
+    // Stores the option string and value in the tagSurrogates hash map and returns a surrogate key.
+    private String getSurrogateKeyForTaggedOption(
+            final String rawOptionString,   // the raw option string provided by the user, including the prefix and option name
+            final String rawArgumentValue,  // the raw argument value provided by the user
+            final String tagString          // the tag string that has been peeled off of the raw option string, including logical name and any attributes
+    )
+    {
+        final String surrogateKey = makeSurrogateKey(rawOptionString, rawArgumentValue);
+        if (null != tagSurrogates.put(surrogateKey, Pair.of(tagString, rawArgumentValue))) {
+            throw new CommandLineException.BadArgumentValue(
+                    String.format("The argument value: \"%s %s\" was duplicated on the command line", rawOptionString, rawArgumentValue));
+        }
+        return surrogateKey;
+    }
+
+    /**
+     * Construct a surrogate key from the option name/tag string (as provided by the user) and raw argument value (as
+     * provided by the user). In order to ensure uniqueness of the key, it needs to include all elements from the
+     * command line, including:
+     *
+     *  -the option name (short/long) used
+     *  -the tag string used
+     *  -the argument value used
+     *
+     * @param rawOptionString the raw option string as provided by the user, without the option prefix
+     * @param rawArgumentValue the entire raw argument value provided by the user
+     * @return a surrogate key that can be included as a substitute value in the command line args presented to
+     * command line the parser
+     */
+    private String makeSurrogateKey(
+            final String rawOptionString,
+            final String rawArgumentValue
+    ) {
+        // The keys strings are never parsed, but the separator is added to the
+        // middle to make visual key inspection easier
+        return rawOptionString + ARGUMENT_TAG_NAME_SEPARATOR + rawArgumentValue;
+    }
+
+    /**
+     * Parse a tag string and populate a TaggedArgument with values.
+     *
+     * @param taggedArg TaggedArgument to receive tags
+     * @param longArgName name of the argument being tagged
+     * @param tagString tag string (including logical name and attributes, no option name)
+     */
+    public void populateArgumentTags(final TaggedArgument taggedArg, final String longArgName, final String tagString) {
+        final ParsedArgument pa = ParsedArgument.of(longArgName, tagString);
+        taggedArg.setTag(pa.getName());
+        taggedArg.setTagAttributes(pa.keyValueMap());
+    }
+
+    /**
+     * Given a TaggedArgument implementation and along argument name, return a string representation of argument,
+     * including the tag and attributes, for display purposes.
+     *
+     * @param taggedArg implementation of TaggedArgument interface
+     * @return a display string representing the tag name and attributes
+     */
+    public static String getDisplayString(final String longArgName, final TaggedArgument taggedArg) {
+        Utils.nonNull(longArgName);
+        Utils.nonNull(taggedArg);
+
+        StringBuilder sb = new StringBuilder();
+        sb.append(longArgName);
+        sb.append(ARGUMENT_TAG_NAME_SEPARATOR);
+        sb.append(taggedArg.getTag());
+
+        if (taggedArg.getTagAttributes() != null) {
+            taggedArg.getTagAttributes().entrySet().stream()
+                    .forEach(
+                            entry -> {
+                                sb.append(ARGUMENT_KEY_VALUE_PAIR_DELIMITER);
+                                sb.append(entry.getKey().toString());
+                                sb.append(ARGUMENT_KEY_VALUE_SEPARATOR);
+                                sb.append(entry.getValue().toString());
+                            });
+        }
+        return sb.toString();
+    }
+
+    /**
+     * Represents a parsed, tagged argument.
+     *
+     * May have attributes.
+     */
+    private static final class ParsedArgument{
+        private final String name;
+        private final Map<String, String> keyValueMap;
+
+        /**
+         * Parses an argument value String of the forms:
+         *
+         * "logical_name(,key=value)*"
+         *
+         * into logical name and key=value pairs.
+         *
+         * @param rawTagValue tag string value from the command line (does not include the argument name) to parse
+         * @return The argument parsed from the provided string.
+         */
+        public static ParsedArgument of(final String longArgName, final String rawTagValue) {
+            final String[] tokens = rawTagValue.split(ARGUMENT_KEY_VALUE_PAIR_DELIMITER, -1);
+
+            if (tokens.length == 0) {
+                throw new CommandLineException.BadArgumentValue(longArgName, rawTagValue, USAGE);
+            }
+            // first token is required to be a name
+            if (tokens[0].contains(ARGUMENT_KEY_VALUE_SEPARATOR)) {
+                throw new CommandLineException.BadArgumentValue("Missing tag name for argument: " + rawTagValue);
+            }
+            if ( Arrays.stream(tokens).anyMatch(String::isEmpty)) {
+                throw new CommandLineException.BadArgumentValue(longArgName, rawTagValue, "Empty tag or attribute encountered. " + USAGE);
+            }
+
+            final ParsedArgument pa = new ParsedArgument(tokens[0]);
+            if (tokens.length == 1) {
+                return pa;
+            } else {
+                // User specified a logical name (and optional list of key-value pairs)
+                for (int i = 1; i < tokens.length; i++){
+                    final String[] kv = tokens[i].split(ARGUMENT_KEY_VALUE_SEPARATOR, -1);
+                    if (kv.length != 2 || kv[0].isEmpty() || kv[1].isEmpty()){
+                        throw new CommandLineException.BadArgumentValue("", rawTagValue, USAGE);
+                    }
+                    if (pa.containsKey(kv[0])){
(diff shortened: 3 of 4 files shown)


cmnbroad:
There is no integration with docgen or usage. I made a separate [ticket](https://github.com/broadinstitute/barclay/issues/35) for that.

cmnbroad:
@droazen BTW, by the time parsing is done, the CLP pretty much has an implementation of TaggedArgument for a given argument (ish - its not a direct implementation right now but it could be). It would be a pretty minor code change to have the CLP hand a tear-off implementation directly to the field via something like a TaggedArgumentConsumer interface.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgument.java
```diff
@@ -0,0 +1,36 @@
...
+     * Set the (optional) tag name for this instance.
+     * @param tagName
+     */
+    void setTag(String tagName);
+
+    /**
+     * Retrieve the tag name for this instance.
+     * @return String representing the tagName, or null if no name was set.
+     */
+    String getTag();
```
cmnbroad: I just noticed that the javadoc comments here are inconsistent with the rest of the implementation, which requires a tag name. The javadoc here should change.

cmnbroad:
@droazen In looking at getCommandLine, I noticed that tagged args need to be treated specially there in order to ensure the tags are included in the output. Since I can't tell if you've started reviewing this, I'm pushing another commit to include that since I assume we'll want it in this release of Barclay. Let me know if you want me to squash.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
```diff
@@ -0,0 +1,333 @@
...
+ *
+ *     --argument_name:logical_name argument_value
+ *
+ * or:
+ *
+ *     --argument_name:logical_name,key1=value1,key2=value2 argument_value
+ *
+ * The logical name is required; key/value pairs are optional. @Argument annotated fields that require tagged
+ * values must implement {@link TaggedArgument}, and the command line parser will require tag names for any such
+ * fields.
```
droazen: Whoops, I think we misunderstood each other the other day when we were talking about whether tags should be required or not. It's not the case that arguments that implement `TaggedArgument` *must* have a tag. What I was saying was that if attributes are present, then a tag name must also be present, but it should be perfectly legal for a `TaggedArgument` field to have neither a tag nor attributes.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
```diff
@@ -0,0 +1,333 @@
...
+ * name, and tag string), and the raw argument value together to ensure the key is unique:
+ *
+ *      Key: --argument_name:logical_name,key1=value1,key2=value2:argument_value
+ *
+ * The pair object stored in the map contains only the (previously peeled off) tag string, and the raw argument value that
+ * was provided by the user. The command line parser replaces the original arguments provided by the user with the option
+ * name and the key:
+ *
+ *          User provides:              "--argument_name:logical_name,key1=value1,key2=value2" "raw_argument_value"
+ *          Parser is presented with:   "--argument_name" "--argument_name:logical_name,key1=value1,key2=value2:argument_value"
```
droazen: The parser is presented with an argument value that begins with `--`? Doesn't this causing parsing ambiguities/issues?

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
```diff
@@ -0,0 +1,333 @@
...
+
+    /**
+     * Separator between keys and values in the "logical_name,key1=value1,key2=value2" syntax.
+     */
+    private static final String ARGUMENT_KEY_VALUE_SEPARATOR = "=";
+
+    /**
+     * Separator used between option name and logical name.
+     */
+    private static final char ARGUMENT_TAG_NAME_SEPARATOR = ':';
```
droazen: Is there a reason only this constant is typed as `char`? Can they all be `chars` for faster splitting?

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
```diff
@@ -0,0 +1,333 @@
...
+
+    private static final String USAGE = "Argument must of the form argument_name:logical_name(,key=value)*";
```
droazen: "must be of the form argument_name or argument_name:logical_name(,key=value)*"

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
```diff
@@ -0,0 +1,333 @@
...
+     * values.
+     * @param args raw arguments as provided by the user
+     * @return curated string of arguments to be presented to the opt parser
+     */
+    public String[] processTaggedOptions(final String[] args) {
+        List<String> finalArgs = new ArrayList<>(args.length);
+        for (int i = 0; i < args.length; ) {
+            String arg = args[i];
+            if (isShortOptionToken(arg)) {
+                i += replaceTaggedOption(SHORT_OPTION_PREFIX, arg.substring(SHORT_OPTION_PREFIX.length()), args, finalArgs, i);
```
droazen: Manipulating a loop counter within a for loop is usually a bad idea/error prone -- I think you could avoid the need for this by converting from an array to an iterator of Strings.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
```diff
@@ -0,0 +1,333 @@
...
+    private static final char ARGUMENT_TAG_NAME_SEPARATOR = ':';
+
+    private static final String USAGE = "Argument must of the form argument_name:logical_name(,key=value)*";
+
+    // Map of surrogate keys to Pair(tag_string, raw_argument_value)
+    private Map<String, Pair<String, String>> tagSurrogates = new HashMap<>();
+
+    // Prefixes used by the opt parser for short/long prefixes
+    private static final String SHORT_OPTION_PREFIX = "-";
+    private static final String LONG_OPTION_PREFIX = "--";
```
droazen: These seem like they should be barclay-wide constants.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
cmnbroad: Ok.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
cmnbroad: Hm. I'm not defining something new here - those are just the switches that are built into the opt parser. The comment is just an illustration of what happens to the command line.

droazen: Are you replacing the argument *values* for tagged arguments in the preprocessed command line with tokens that begin with `--` or `-`? If yes, that seems like a bad idea.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
```diff
@@ -0,0 +1,333 @@
...
+     * curated key and value. Otherwise just add the raw option.
+     *
+     * @param optionPrefix the actual option prefix used for this option, either "-" or "--"
+     * @param optionString the option string including everything but the prefix
+     * @param userArgList the raw arguments provided by the user, used to retrieve the value corresponding
+     *                    to this option
+     * @param finalArgList the curated argument list
+     * @param i the index into the raw arg list of the option we're currently processing
+     * @return the number of items on the user arg list consumed by this method. This will either 1 if this
+     * option is not tagged) or 2 (if the option is tagged)
```
droazen: Would be cleaner if this took an `Iterator<String>` for the user args and returned a `List<String>` tokens to be added to the final args. That way you don't need to pass around indices and pass in lists to be modified.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
cmnbroad: Ok, but I would think we want to discourage using these anywhere else. Its kind of gross that I even have to do this at all. If I could access the ones used by the opt parser I would do that instead.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
cmnbroad: Ah, I see what you're saying. I originally wasn't including the prefix in the key, but added it to disambiguate the short and long name cases. But you're right its probably safer to move it so its not at the beginning of the key string.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
```diff
@@ -0,0 +1,333 @@
...
+     * @return a surrogate key that can be included as a substitute value in the command line args presented to
+     * command line the parser
+     */
+    private String makeSurrogateKey(
+            final String rawOptionString,
+            final String rawArgumentValue
+    ) {
+        // The keys strings are never parsed, but the separator is added to the
+        // middle to make visual key inspection easier
+        return rawOptionString + ARGUMENT_TAG_NAME_SEPARATOR + rawArgumentValue;
```
droazen: Concatenating the argument value onto the argument name/tags like this seems very problematic to me -- you now create the same parsing ambiguities that we had before with argument values like `hostname:port/path`. 

Can you find a way to avoid doing this concatenation?

cmnbroad: The concatenated string is never parsed; its jut a unique key used to retrieve the values as they originally appeared on the command line.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
droazen: I'm not convinced you're adding it, actually -- it looks like it gets stripped before the key is made.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
droazen: Never mind, I think I spoke too soon -- I think you avoid the parsing ambiguity by never parsing the surrogate key, but I need to read a bit more.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
cmnbroad: That one is a char because its used with indexOf. The others are used with split and contains, which take a String.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
```diff
@@ -0,0 +1,333 @@
...
+        }
+        return surrogateKey;
+    }
+
+    /**
+     * Construct a surrogate key from a raw option string (as provided by the user) and raw argument value (as
+     * provided by the user). In order to ensure uniqueness of the key, it needs to include all elements from the
+     * command line, including:
+     *
+     *  -the option prefix (short/long) used
```
droazen: Does the key actually start with the `--`/`-` prefix? I think it's stripped based on my reading of the code, but if it's not stripped it should be sanitized to avoid confusing the parser with an argument value that looks like an argument name.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
cmnbroad: Just checked in the debugger - its definitely not there. I also just verified that you can't use a shortName from one arg as the longName for another. So I think its fine. Its the comments that need to change.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
droazen: Yes, confirmed that you are fine, since you rely on the `Pair` stored in the map for separation of the tag from the value, and don't parse the key. Disregard this comment chain.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
droazen: Up to you if you want to hide them out of shame :)

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
```diff
@@ -0,0 +1,333 @@
...
+            } else {
+                // the option appears to be tagged, but we're already at the end of the argument list,
+                // and there is no companion value to use
+                throw new CommandLineException("No value found for tagged argument: " + optionString);
+            }
+        }
+    }
+
+    /**
+     * Attempt to retrieve a
```
droazen: Method description is truncated

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
```diff
@@ -0,0 +1,333 @@
...
+                && !SHORT_OPTION_PREFIX.equals( argument )
+                && !isLongOptionToken( argument );
+    }
+
+    // See if the opt parser would think this is a long option ("--")
+    private static boolean isLongOptionToken(final String argument) {
+        return argument.startsWith( LONG_OPTION_PREFIX );
+    }
+
+    private String getSurrogateKeyForTaggedOption(
```
droazen: Document that this method also updates `tagSurrogates`

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
```diff
@@ -0,0 +1,333 @@
...
+    private static final String LONG_OPTION_PREFIX = "--";
+
+    /**
+     * Given an array of raw arguments provided by the user, return an array of args where tagged arguments
+     * have been replaced with curated arguments containing a key to be used by the parser to retrieve the actual
+     * values.
+     * @param args raw arguments as provided by the user
+     * @return curated string of arguments to be presented to the opt parser
+     */
+    public String[] processTaggedOptions(final String[] args) {
```
droazen: process -> preprocess

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
```diff
@@ -0,0 +1,333 @@
...
+    ) {
+        // The keys strings are never parsed, but the separator is added to the
+        // middle to make visual key inspection easier
+        return rawOptionString + ARGUMENT_TAG_NAME_SEPARATOR + rawArgumentValue;
+    }
+
+    /**
+     * Parse a tag string and populate a TaggedArgument with values.
+     * @param tagString tag string (including logical name and attributes, no option name)
+     * @param taggedArg TaggableArgument to receive tags
```
droazen: `TaggableArgument` -> `TaggedArgument`

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
```diff
@@ -0,0 +1,333 @@
...
+     *
+     * @param taggedArg implementation of TaggedArgument interface
+     * @return a display string representing the tag name and attributes
+     */
+    public static String getDisplayString(final TaggedArgument taggedArg) {
+        Utils.nonNull(taggedArg);
+
+        StringBuilder sb = new StringBuilder();
+        sb.append(ARGUMENT_TAG_NAME_SEPARATOR);
+        sb.append(taggedArg.getTag() == null ? "NoTagName" : taggedArg.getTag());
```
droazen: If a tagged argument has no tag, its display string should be simply the argument name. The case `getTag() == null && getTagAttributes() != null` should be disallowed, so you don't have to worry about it here.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
```diff
@@ -0,0 +1,333 @@
...
+
+        // check for nulls since we can't assume the implementation is compliant
+        if (taggedArg.getTagAttributes() != null) {
+            taggedArg.getTagAttributes().entrySet().stream()
+                    .filter(entry -> entry.getValue() != null)
```
droazen: An attribute with a null value should not be allowed by the parser.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
```diff
@@ -0,0 +1,333 @@
...
+     * May have attributes.
+     */
+    private static final class ParsedArgument{
+        private final String name;
+        private final Map<String, String> keyValueMap;
+
+        /**
+         * Parses an argument value String of the forms:
+         *
+         * "logical_name(,key=value)"
```
droazen: `logical_name(,key=value)` -> `logical_name(,key=value)*`

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
```diff
@@ -0,0 +1,333 @@
...
+         *
+         * into logical name and key=value pairs.
+         *
+         * @param rawTagValue argument value from the command line to parse
```
droazen: Make it clear that this String doesn't include the argument name itself.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
```diff
@@ -0,0 +1,333 @@
...
+         * @return The argument parsed from the provided string.
+         */
+        public static ParsedArgument of(final String rawTagValue) {
+            final String[] tokens = rawTagValue.split(ARGUMENT_KEY_VALUE_PAIR_DELIMITER);
```
droazen: Use `rawTagValue.split(ARGUMENT_KEY_VALUE_PAIR_DELIMITER, -1)` so that empty tokens (eg., `,,`) can be detected and raise an error. From the docs for `split()`:

```
<p> The {@code limit} parameter controls the number of times the
     * pattern is applied and therefore affects the length of the resulting
     * array.  If the limit <i>n</i> is greater than zero then the pattern
     * will be applied at most <i>n</i>&nbsp;-&nbsp;1 times, the array's
     * length will be no greater than <i>n</i>, and the array's last entry
     * will contain all input beyond the last matched delimiter.  If <i>n</i>
     * is non-positive then the pattern will be applied as many times as
     * possible and the array can have any length.  If <i>n</i> is zero then
     * the pattern will be applied as many times as possible, the array can
     * have any length, and trailing empty strings will be discarded.
```

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
```diff
@@ -0,0 +1,333 @@
...
+            if ( Arrays.stream(tokens).anyMatch(String::isEmpty)) {
+                throw new CommandLineException.BadArgumentValue("", rawTagValue, "Empty tag name encountered. " + USAGE);
```
droazen: It's not necessarily an empty tag name -- it's either an empty tag or an empty key/value pair.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
```diff
@@ -0,0 +1,333 @@
...
+            }
+            if (tokens.length == 0) {
+                throw new CommandLineException.BadArgumentValue("", rawTagValue, USAGE);
```
droazen: Need an error message here

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
```diff
@@ -0,0 +1,333 @@
...
+            // first token is required to be a name
+            if (tokens[0].contains(ARGUMENT_KEY_VALUE_SEPARATOR)) {
+                throw new CommandLineException.BadArgumentValue("Missing name for argument: " + rawTagValue);
+            }
+            final ParsedArgument pa = new ParsedArgument(tokens[0]);
+
+            if (tokens.length == 1) {
+                // even though its parseable, notify the user for the case where an attribute is indicated, but not
+                // included, such as for a dangling comma ("--t:tumor,") since it may be user error
+                if (rawTagValue.contains(ARGUMENT_KEY_VALUE_PAIR_DELIMITER)) {
```
droazen: If you call split with `-1` above, you should be able to delete this check, since split will not drop the empty tokens and they'll be caught by the `isEmpty` check above.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
cmnbroad: Its not, but TaggedArgument is an interface. The parser doesn't own the implementation.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgument.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Interface for arguments that can have tags and attributes. The command line argument parser
+ * looks for argument fields that implement this interface, and propagates tag names and
+ * attribute/value pairs to the field.
+ */
+public interface TaggedArgument {
+
+    /**
+     * Set the tag name (required) for this instance.
```
droazen: optional -- required only if attributes are also present

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -615,7 +619,37 @@ private void setArgument(ArgumentDefinition argumentDefinition, final List<Strin
...
                 }
             } else {
-                value = constructFromString(CommandLineParser.getUnderlyingType(argumentDefinition.field), stringValue, argumentDefinition.getLongName());
+                // See if the value is a surrogate key in the tag parser's map that was placed there during preprocessing,
+                // and if so, unpack the values retrieved via the key and use those to populate the field
+                Pair<String, String> taggedOptionPair = tagParser.getTaggedOptionForSurrogate(stringValue);
+                if (TaggedArgument.class.isAssignableFrom(argumentDefinition.type)) {
+                    if (taggedOptionPair == null) {
+                        throw new CommandLineException(
+                                "A tag name is required for the argument: " + argumentDefinition.shortName + ":" + argumentDefinition.fullName);
```
droazen: Should be allowed, as discussed earlier

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -615,7 +619,37 @@ private void setArgument(ArgumentDefinition argumentDefinition, final List<Strin
...
+                                taggedOptionPair.getRight(),        // argument value
+                                argumentDefinition.getLongName());
+                        TaggedArgument taggedArgument = (TaggedArgument) value;
+                        tagParser.populateArgumentTags(taggedArgument, taggedOptionPair.getLeft());
+                    }
+                }
+                else {
+                    if (taggedOptionPair != null) { // a tag was found for a non-taggable argument
+                        throw new CommandLineException(
+                                String.format("The argument: \"%s/%s\" does not accept tags: \"%s\"",
```
droazen: This should remain an error

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -1149,7 +1183,15 @@ private String prettyNameValue(Object value) {
                 if (isSensitive){
                     return String.format("--%s ***********", getLongName());
                 } else {
-                    return String.format("--%s %s", getLongName(), value);
+                    if (value instanceof TaggedArgument) {
+                        TaggedArgument taggedArg = (TaggedArgument) value;
+                        return String.format("--%s%s %s",
+                                getLongName(),
+                                TaggedArgumentParser.getDisplayString(taggedArg),
```
droazen: Tagged arguments without tags should display as just `--longName value`

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
```diff
@@ -0,0 +1,301 @@
...
+
+/**
+ * Tests for taggable arguments, i.e.:
+ *
+ *     -I:tumor my.bam
+ *     -I:tumor,key=value,key2=value2
+ */
+public class TaggedArgumentTest {
+
+    public static class TaggableArg implements TaggedArgument {
```
droazen: `private`

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
```diff
@@ -0,0 +1,301 @@
...
+        @Override
+        public String toString() { return argValue; }
+    }
+
+    @CommandLineProgramProperties(
+            summary = "Test tagged aruments",
+            oneLineSummary = "Test tagged aruments",
+            programGroup = TestProgramGroup.class
+    )
+    public class TaggableArguments {
```
droazen: `private`, if possible

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
```diff
@@ -0,0 +1,301 @@
...
+
+        @PositionalArguments(minElements = 0, maxElements=2)
+        public List<String> positionalArgs = new ArrayList<>();
+
+        @Argument(shortName="t", fullName="tFullName", doc="Tag target arg", optional = true)
+        public List<TaggableArg> taggableArgList = new ArrayList<>();
+
+        @Argument(shortName="s", fullName="scalarArg", optional = true)
+        public Integer taggableInteger = 17;
```
droazen: How is this taggable? `Integer` does not implement `TaggedArgument`?

(We should have a test for a taggable non-Collection argument, though)

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
```diff
@@ -0,0 +1,301 @@
...
+
+        @Argument(shortName="n", fullName="notTaggable", optional = true)
+        public String notTaggable;
+    }
+
+    // Non-generic key-value class for easy addition to statically initialized arrays in data provider
+    private static class KeyValuePair{
```
droazen: Can't you use `org.apache.commons.lang3.tuple.Pair`, which also has a static `of()` method?

cmnbroad: Its generic - you can have an array of generics.

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    @SuppressWarnings("unchecked")
+    @DataProvider(name = "GoodTaggedArguments")
+    public Object[][] goodTaggedArguments() {
+        return new Object[][][]{
+                //TODO: file args, tagged scalar/enum (?), intermix tagged with positional args
+                {
+                        // single value in a collection, no attributes, short name
+                        new String[]{"--t:tumor", "tumor.bam"},
```
droazen: Need some valid test cases with short-name arguments as well, since you manually check for `--` vs. `-` in your parsing code.

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
cmnbroad: Its just a poorly named argument. I wanted to intermix taggeble and not, so I added to have something that isn't taggable, and chose a name too hurriedly.

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
```diff
@@ -0,0 +1,301 @@
...
+                },
+                {
+                        // single value in a collection, no attributes, short name
+                        new String[]{"--t:tumor", "/tumor.bam"},
+                        new KeyValuePair[]{KeyValuePair.of("tumor", "/tumor.bam")},
+                        new KeyValuePair[][]{{}}
+                },
+                {
+                        // single tagged value in a collection, no attributes, short name
+                        new String[]{"--t:tumor", "gcs://my/tumor.bam"},
```
droazen: Add additional test cases: `"--t:tumor", "hostName:1234/tumor.bam"` and `"--t:tumor", "http://hostName:1234/tumor.bam"`

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
droazen: Wait, `t` *is* the short name for this arg, yet you are prefixing it with `--`? This doesn't seem right.

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
```diff
@@ -0,0 +1,301 @@
...
+                },
+                {
+                        // single value in a collection, no attributes, full name
+                        new String[]{"--tFullName:tumor", "tumor.bam"},
+                        new KeyValuePair[]{KeyValuePair.of("tumor", "tumor.bam")},
+                        new KeyValuePair[][]{{}}
+                },
+                {
+                        // single tagged value in a collection, two attributes, short name
+                        new String[]{"--t:tumor,truth=true,training=false", "tumor.bam"},
```
droazen: Also add a test case for a tag with just a single key/value pair

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
```diff
@@ -0,0 +1,301 @@
...
+                        new KeyValuePair[][]{
+                                {
+                                        KeyValuePair.of("truth", "false"),
+                                        KeyValuePair.of("training", "true")
+                                }
+                        }
+                },
+                {
+                        // two tagged values in a collection, no attributes, short name
+                        new String[]{"--t:tumor", "tumor.bam", "--t:normal", "normal.bam"},
```
droazen: Add test case: two args, one tagged, one untagged, and assert that the untagged one does not get tagged

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
```diff
@@ -0,0 +1,301 @@
...
+                        }
+                }
+        };
+    }
+
+    @Test(dataProvider="GoodTaggedArguments")
+    public void testGoodTaggedArguments(
+            final String argv[],
+            final KeyValuePair tagNameValuePairs[],
+            final KeyValuePair attributePairArrays[][])
```
droazen: `tagNameValuePairs` -> `expectedTagNameValuePairs`
`attributePairArrays` -> `expectedAttributePairArrays`

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
```diff
@@ -0,0 +1,301 @@
...
+            Map<String, String> attributes = taggable.taggableArgList.get(i++).getTagAttributes();
+            for (KeyValuePair attributePair : attributePairArray) {
+                Assert.assertEquals(attributes.get(attributePair.getKey()), attributePair.getValue());
+            }
+        }
+    }
+
+    @DataProvider(name = "BadTaggedArguments")
+    public Object[][] badTaggedArguments() {
+        return new Object[][]{
```
droazen: Here too you need some test cases with `-` rather than `--`

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
```diff
@@ -0,0 +1,301 @@
...
+                // long name
+                {new String[]{"--t"}},                              // taggable, but no tagname, no arg
```
droazen: Not an error

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
```diff
@@ -0,0 +1,301 @@
...
+                {new String[]{"--t:"}},                             // taggable, tagname missing, no arg
+                {new String[]{"--t:", "tumor1.bam"}},               // taggable, tagname missing, with arg
+                {new String[]{"--t:tagName"}},                      // taggable, with tagname, but no arg
+
+                // short name
```
droazen: full name

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
```diff
@@ -0,0 +1,301 @@
...
+            for (KeyValuePair attributePair : attributePairArray) {
+                Assert.assertEquals(attributes.get(attributePair.getKey()), attributePair.getValue());
+            }
+        }
+    }
+
+    @DataProvider(name = "BadTaggedArguments")
+    public Object[][] badTaggedArguments() {
+        return new Object[][]{
+                // long name
```
droazen: short name

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
```diff
@@ -0,0 +1,301 @@
...
+                // missing value
+                {new String[]{"--t:tagName,key=value"}},             // taggable, with tagname, attributes, missing value at end
+                {new String[]{"--t:tagName,key=value", "-s", "28"}}, // taggable, with tagname, attributes, missing value and second argument
+                {new String[]{ "-s", "28", "--t:tagName,key=value"}},// second argument taggable with tagname, attributes, missing value
+
+                // Malformed attribute strings
+                {new String[]{"--t:tumor,truth", "tumor.bam"}},                    // attribute name with missing value
+                {new String[]{"--t:tumor,truth=", "tumor.bam"}},                   // attribute name with missing value
+                {new String[]{"--t:tumor,truth=true,truth=false", "tumor.bam"}},   // duplicate attribute value
+                {new String[]{"--t:tumor,", "tumor.bam"}},                         // dangling comma
```
droazen: Need a lot of additional test cases in this `DataProvider`:

`--arg:tag:tag value`
`--arg:tag,key=value=foo value`
`--arg:tag,, value`
`--arg:tag,,key=value value`
`--arg:,key=value value`
`--arg:,,key=value value`

etc. etc. See `InvalidFeatureArgumentValuesDataProvider` in `FeatureInputUnitTest` in GATK4, and try to replicate all the test cases there.

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
```diff
@@ -0,0 +1,301 @@
...
+
+                // TODO: CLP doesn't have state that easily supports checking for this - leave this to the tool ?
+                //{new String[]{"--t:tumor", "tumor1.bam", "--t:tumor", "tumor2.bam"}} // duplicate tagname in a collection
```
droazen: Fine to leave this to the tool (but document in the `TaggedArgument` interface that this check must be done by the client).

droazen:
Review complete, back to @cmnbroad. Looks good overall, please merge after addressing my comments, and update the barclay snapshot in GATK4 so that it can use this feature.

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
droazen: Ok, but can you add a non-Collection field of type `TaggableArgument`, and test cases to exercise it?

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
cmnbroad: Done.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
cmnbroad: Done.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
cmnbroad: No. Fixed.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
cmnbroad: Done.

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
cmnbroad: Done.

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
cmnbroad: Done.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
cmnbroad: Done.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
cmnbroad: Done.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgument.java
cmnbroad: Done.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
cmnbroad: Done.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
cmnbroad: Done.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
cmnbroad: Done.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
cmnbroad: Done.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
cmnbroad: Done.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
cmnbroad: Done.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
cmnbroad: Done.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
cmnbroad: Done.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
cmnbroad: Removed.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
cmnbroad: Done.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
cmnbroad: Done.

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
cmnbroad: Done.

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
cmnbroad: Done.

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
cmnbroad: There are several tests in here that use the short name, or the long name, and some that mix. I do need to add some tests that use "-", though. As for the parser, I don't believe it cares - you can use either of "-" or "--" with either name.

cmnbroad: Done.

cmnbroad: The parser allows you intermix either.

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
cmnbroad: Done.

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
cmnbroad: Done.

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
cmnbroad: Done.

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
cmnbroad: Done.

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
cmnbroad: Its an error because it has no argument value, not because it has no tag.

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
cmnbroad: Done.

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
cmnbroad: Done.

src/test/java/org/broadinstitute/barclay/argparser/TaggedArgumentTest.java
cmnbroad: Duplicated all with both. Done.

src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
cmnbroad: Switched to iterators and removed indices.

Issue #23, opened 2017-01-18
Primitive types are always optional


magicDGS:
Because they cannot be `null`, they are always treated as optional by the CLP.

cmnbroad:
There are two side-effects of this. First, as mentioned above, arguments that have primitive types are never flagged by the parser as not having been set, even if they're required, because the parser always views them as having a value, even if it was set programmatically.

Second, if affects plugins such as the read filter plugin because the parser presents the argument to the plugin as having been supplied (via a call to isDependentArgumentAllowed) This is problematic when the value was actually supplied as a default value for a default plugin by the tool, but the user want to disable the filter. During validation, the plugin will complain that an argument value was supplied for a filter that is also being disabled, even though the user didn't supply the argument value.

PR #38, opened 2017-02-05
Updates to CommandLinePluginDescriptor interface.

cmnbroad:
DO NOT MERGE.

@magicDGS Please take a look at the changes to the descriptor interface in this branch. I've changed some names, as well as some method contract descriptions.

If the read filter plugin implemented this interface, and you had the ability inject argument definitions as you do in https://github.com/broadinstitute/gatk/pull/2355, would that be sufficient to enable your filter plugin use cases using just method overrides ?

Summary of the main changes I made:

- getAllInstances -> getResolvedInstances (changed contract to return all “resolved” instances, including defaults, and note that resulting list may be different if called pre/post parseArgs)
- getDefaultInstances -> getDefaultInstances (changed contract to say the resulting list may be different if called pre/post parseArgs)
- getPluginClass -> getPluginBaseClass
- getClassFilter -> includePluginClass (simplified to return a boolean rather than a predicate)
- getInstance -> createInstanceForPlugin
- validateArguments -> validateAndResolvePlugins

### DIFF
--- a/src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
+++ b/src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
@@ -2,55 +2,103 @@ package org.broadinstitute.barclay.argparser;
 
 import java.util.List;
 import java.util.Set;
-import java.util.function.Predicate;
 
 /**
- * A base class for descriptors for plugins that can be dynamically discovered by the
- * command line parser and specified as command line arguments. An instance of each
- * plugin descriptor to be used should be passed to the command line parser, and will
- * be queried to find the class and package names to search for all plugin classes
- * that should be discovered dynamically. The command line parser will find all such
- * classes, and delegate to the descriptor to obtain the corresponding plugin instance;
- * the object returned to the parser is then added to the parser's list of argument sources.
- *
- * Descriptors (sub)classes should have at least one @Argument used to accumulate the
- * user-specified instances of the plugin seen on the command line. Allowed values for
- * this argument are the simple class names of the discovered plugin subclasses.
- *
- * Plugin (sub)classes:
- *
- * - should subclass a common base class (the name of which is returned by the descriptor)
- * - may live in any one of the packages returned by the descriptor {@Link #getPackageNames},
- *   but must have a unique simple name to avoid command line name collisions.
- * - should contain @Arguments for any values they wish to collect. @Arguments may be
- *   optional or required. If required, the arguments are in effect "provisionally
- *   required" in that they are contingent on the specific plugin being specified on
- *   the command line; they will only be marked by the command line parser as missing
- *   if the they have not been specified on the command line, and the plugin class
- *   containing the plugin argument *has* been specified on the command line (as
- *   determined by the command line parser via a call to isDependentArgumentAllowed).
- *
- * NOTE: plugin class @Arguments that are marked "optional=false" should be not have a primitive
- * type, and should not have an initial value, as the command line parser will interpret these as
- * having been set even if they have not been specified on the command line. Conversely, @Arguments
- * that are optional=true should have an initial value, since they parser will not require them
- * to be set in the command line.
- *
- * The methods for each descriptor are called in the following order:
- *
- *  getPluginClass()/getPackageNames() - once when argument parsing begins (if the descriptor
- *  has been passed to the command line parser as a target descriptor)
- *
- *  getClassFilter() - once for each plugin subclass found
- *  getInstance() - once for each plugin subclass that isn't filtered out by getClassFilter
- *  validateDependentArgumentAllowed  - once for each plugin argument value that has been
- *  specified on the command line for a plugin that is controlled by this descriptor
- *
- *  validateArguments() - once when argument parsing is complete
- *  getAllInstances() - whenever the pluggable class consumer wants the resulting plugin instances
- *
- *  getAllowedValuesForDescriptorArgument is only called when the command line parser is constructing
- *  a help/usage message.
+ * <p>
+ * Base class for a descriptor that defines a set of command line plugin classes that can be dynamically
+ * discovered by the command line parser. Plugin instances contain argument fields that extend the
+ * set of command line arguments accepted by the parser.
+ * </p>
+ * <p>
+ * Each plugin descriptor passed to the command line parser is queried to determine the class and
+ * package names to be searched for plugin classes controlled by that descriptor. The descriptor
+ * object itself is added to the parser's list of argument sources, and for each matching plugin class
+ * discovered in the packages list defined by the descriptor, the parser delegates back to the descriptor
+ * to obtain an instance of that plugin. Each of these plugin objects are then also added to the parsers's
+ * list of argument sources.
+ * </p>
+ * <p>
+ * Descriptor implementations should have at least one @Argument field used to accumulate the plugins that
+ * are specified on the command line by the user. Allowed values for this argument are taken from the
+ * plugins that have been dynamically discovered.
+ * </p>
+ * <p>
+ * Descriptor implementations should also maintain a list of the instances that are returned to callers by
+ * {@link #createInstanceForPlugin} so they can be returned through {@link #getResolvedInstances} once
+ * validation is complete.
+ * </p>
+ * <p>
+ * A descriptor implementation may optionally choose to provide a constructor that specifies a list of
+ * default plugin instances that are automatically included along with any plugins that are specified
+ * manually by the user. It is the descriptor implementation's job to resolve and validate any command
+ * line arguments with these default instances for purposes of ultimately rendering a fully resolved
+ * list of instances to be returned by {@link #getResolvedInstances}. Implementers should be aware of
+ * https://github.com/broadinstitute/barclay/issues/23 when implementing the resolution policy.</li>
+ * </p>
+ * <p>
+ * Each plugin class controlled by a descriptor:
+ * </p>
+ * <ul>
+ * <li> Must live in one of the packages returned by the {@link #getPackageNames} method. </li>
+ * <li> Must have a type that is a subclass of the plugin base class returned by the {@link #getPluginBaseClass()}
+ *      method </li>
+ * <li> Must have a unique simple class name in order to avoid command line name collisions </li>
+ * <li> May contain one or more @Argument annotated fields for any arguments used with that plugin. Plugin arguments
+ *      may be optional or required. If required, the arguments are actually "conditionally required", in that they
+ *      are only required contingent on the predecessor plugin being specified on the command line, or included in the
+ *      default plugins (as determined by the command line parser via a call to isDependentArgumentAllowed).
+ *      Likewise, the same conditions must apply for any such dependent argument to be accepted as valid by the
+ *      descriptor.</li>
+ * <li> NOTE: plugin class @Arguments that are marked "optional=false" should not have a primitive type, and
+ *      should not have an initial value, as the command line parser will interpret such arguments as having been set even
+ *      if they have not been specified on the command line. Conversely, @Arguments that are optional=true should
+ *      have an initial value, since the parser will not require them to be set in the command line. </li>
+ * </ul>
+ * <p>
+ * The command line parser calls the descriptor methods based on the following descriptor lifecycle:
+ * </p>
+ * <ol>
+ * <li> During command line parser initialization (before plugin discovery or command line parsing begins): </li>
+ * <ul>
+ * <li> {@link #getPackageNames} </li>
+ * <li> {@link #getPluginBaseClass} </li>
+ * </ul>
+ * <li> During plugin discovery (before argument parsing begins). Once for each plugin class discovered: </li>
+ * <ul>
+ * <li> {@link #includePluginClass} </li>
+ * <li> {@link #createInstanceForPlugin} </li>
+ * </ul>
+ * <li> During command line argument parsing: </li>
+ * <ul>
+ * <li> {@link #isDependentArgumentAllowed} - once for each argument that has been seen on the command line that is
+ *      contained within a plugin class that is controlled by this descriptor. The descriptor should decide whether
+ *      the dependent is allowed argument by determining if the predecessor class is valid (implementations should
+ *      recognize that the predecessor class may have been provided as a default plugin, even if not specified by
+ *      the user on the command line) usually based on what plugin name arguments and default instances have been
+ *      provided for this descriptor.
+ * </li>
+ * </ul>
+ * <li> When command line argument parsing is complete (all command line arguments have been processed), but
+ * before the parser returns control to the caller, a one time call: </li>
+ * <ul>
+ * <li> {@link #validateAndResolvePlugins} </li>
+ * </ul>
+ * <li> After the command line parser returns from argument parsing. These methods are generally called by
+ * the application that consumes the plugins: </li>
+ * <ul>
+ * <li> {@link #getDefaultInstances} - whenever a plugin consumer wants the list of plugin instances that were
+ *      specified as defaults. Note that is possible that the list returned by this method will be different
+ *      after a call to {@link #validateAndResolvePlugins}, since the resolution policy may allow the user to
+ *      override and disable default plugins.</li>
+ * <li> {@link #getResolvedInstances} - whenever a plugin consumer wants the final list of resolved of plugin
+ *      instances (including defaults) that were specified on the command line, or otherwise result from the
+ *      the resolution policy determined by {@link #validateAndResolvePlugins}.
+ * </ul>
+ * <li> When generating command line help and during documentation generation: </li>
+ * <ul>
+ * <li> {@link #getAllowedValuesForDescriptorHelp} - only called when the command line parser is constructing
+ *      a help/usage message for an @Argument field </li>
+ * </ul>
  */
 public abstract class CommandLinePluginDescriptor<T> {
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -2,70 +2,101 @@
...
+ * <li> When command line argument parsing is complete (all command line arguments have been processed), but
+ * before the parser returns control to the caller, a one time call: </li>
+ * <ul>
+ * <li> {@link #validateAndResolvePlugins} </li>
+ * </ul>
+ * <li> After the command line parser returns from argument parsing. These methods are generally called by
+ * the application that consumes the plugins: </li>
+ * <ul>
+ * <li> {@link #getDefaultInstances} - whenever a plugin consumer wants the list of plugin instances that were
+ *      specified as defaults </li>
```
magicDGS: As specified in the javadoc for this method, this is dependent on the resolution policy too, so it won't always return the list of plugins specified as defaults. Maybe a small note here will be useful.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -76,103 +107,126 @@
...
-     * help/doc generation.
+     * @return the default plugins enabled for this descriptor as a list of Object. Used for
+     * help/documentation generation for this descriptor.
+     *
+     * NOTE: calling this method after argument parsing (and thus after {@link #validateAndResolvePlugins}
+     * is called) may return a different list than calling it before parsing, since the resolution
+     * policy is implementation-dependent and may depend on the actual argument specified by the user
+     * on the command line.
      */
     public abstract List<Object> getDefaultInstances();
```
magicDGS: I guess that this should be bound too (as in #32).

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -2,70 +2,101 @@
...
+ * <ul>
+ * <li> {@link #validateAndResolvePlugins} </li>
+ * </ul>
+ * <li> After the command line parser returns from argument parsing. These methods are generally called by
+ * the application that consumes the plugins: </li>
+ * <ul>
+ * <li> {@link #getDefaultInstances} - whenever a plugin consumer wants the list of plugin instances that were
+ *      specified as defaults </li>
+ * <li> {@link #getResolvedInstances} - whenever a plugin consumer wants the list of plugin instances that were
+ *      specified on the command line and result from the past-parsing validation performed by
```
magicDGS: As the method's javadoc points out, it also include defaults (even if they are not specified in the command line). Maybe this should be changed from `... specified on the command line and ...` to `... specified on the command line (including defaults) and ...`

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -76,103 +107,126 @@
...
+     * on the command line.
+     */
+    public abstract List<T> getResolvedInstances();
+
+    /**
+     * Return the allowable values for the String argument of this plugin descriptor
+     * that is specified by {@code longArgName}. Used for help/documentation generation for this descriptor.
+     . If the value is unrecognized
+     * the implementation should throw
+     * IllegalArgumentException.
```
magicDGS: Nitpicking comment: this lines are weirdly formatted.

cmnbroad:
@magicDGS Great, and thanks for the comments. And yes, I didn't include the bounds changes in this, but it might be easier for me to just include it here. I'll make another pass and integrate those.

magicDGS:
Any progress on this @cmnbroad? I will appreciate if the first release of GATK already includes their plugins with a consistent interface.

Thank you very much in advance!

cmnbroad:
@asmirnov239 this is the PR mentioned in our discussion abut read filters. Will try to remember to ping you again when this makes its way into GATK.

magicDGS:
Any advance on this, @cmnbroad? Thank you very much in advance!

cmnbroad:
@magicDGS We definitely want to do this for release, but we won't be able to do it for beta.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -2,70 +2,101 @@
...
+ * <ul>
+ * <li> {@link #includePluginClass} </li>
+ * <li> {@link #createInstanceForPlugin} </li>
+ * </ul>
+ * <li> During command line argument parsing: </li>
+ * <ul>
+ * <li> {@link #isDependentArgumentAllowed} - once for each argument that has been seen on the command line that is
+ *      contained within an plugin class that is controlled by this descriptor. The descriptor should decide whether
+ *      the dependent is allowed argument by determining if the predecessor class is valid, usually based on what
+ *      plugin name arguments and default instances have been provided for this descriptor.
```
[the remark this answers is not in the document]
cmnbroad: Done.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
cmnbroad: Done.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
cmnbroad: Done.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -2,70 +2,101 @@
...
+ *      specified as defaults </li>
+ * <li> {@link #getResolvedInstances} - whenever a plugin consumer wants the list of plugin instances that were
+ *      specified on the command line and result from the past-parsing validation performed by
+ *      {@code validateArguments}. </li>
+ * </ul>
+ * <li> Methods  used for command line help and documentation generation: </li>
+ * <ul>
+ * <li> {@link #getAllowedValuesForDescriptorHelp} - only called when the command line parser is constructing
+ *      a help/usage message for an @Argument field </li>
+ * </ul>
```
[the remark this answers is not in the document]
cmnbroad: Yeah. Done.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -76,103 +107,126 @@
...
+    public abstract Class<?> getPluginBaseClass();
+
+    /**
+     * Determine if a plugin class should be retained for this descriptor. Return true if the
+     * descriptor wants this class to be included in from the list of plugins discovered dynamically.
+     *
+     * @return false if the plugin class shouldn't be used; otherwise true
      */
-    public Predicate<Class<?>> getClassFilter() { return c -> true;}
+    public boolean includePluginClass(Class<?> c) { return true;}
```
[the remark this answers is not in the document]
cmnbroad: I'm reluctant to make that change. Unless there is a bug somewhere, isAssignableFrom should never return false. Changing the implementation to call it would silently mask such an issue.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
cmnbroad: Yeah. Fixed.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -76,103 +107,126 @@
...
 
     /**
-     * Give this descriptor a chance to filter out any classes it doesn't want to be
-     * dynamically discoverable.
-     * @return false if the class shouldn't be used; otherwise true
+     * Base class for all command line plugin classes managed by this descriptor. Subclasses of
+     * this class in any of the packages returned by {@link #getPackageNames} will be command line
+     * accessible.
+     */
+    public abstract Class<?> getPluginBaseClass();
```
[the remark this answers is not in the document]
cmnbroad: I'm going to add a separate commit on top to this PR with the bounds changes.

cmnbroad:
@magicDGS I'm reviving this and have made some updates. I'm going to close the old PR with type bounds changes, as I've included those here. There are fewer type bounds changes that we originally had in your old PR though - I've left all of the Class types with unbounded wildcards for consistency and simplicity. Since they can't all be tightened, I'd rather keep them all consistent with each other.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -76,103 +110,127 @@
...
     /**
-     * Give this descriptor a chance to filter out any classes it doesn't want to be
-     * dynamically discoverable.
-     * @return false if the class shouldn't be used; otherwise true
+     * Base class for all command line plugin classes managed by this descriptor. Subclasses of
+     * this class in any of the packages returned by {@link #getPackageNames} will be command line
+     * accessible.
      */
-    public Predicate<Class<?>> getClassFilter() { return c -> true;}
+    public abstract Class<?> getPluginBaseClass();
```
magicDGS: Why not bounding this? In principle, the `CommandLinePluginDescriptor`should be for classes of `T`, no?

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -2,70 +2,104 @@
...
+ * <p>
+ * Descriptor instances should have at least one @Argument field used to accumulate the plugins that
+ * are specified on the command line by the user. Allowed values for this argument are taken from the
+ * plugins that have been dynamically discovered.
+ * </p>
+ * <p>
+ * A descriptor implementation may optionally choose to provide a constructor that specifies a list of
+ * default plugin instances that are automatically included along with any plugins that are specified
+ * manually by the user. The descriptor and command line parser generally treat these default instances
+ * as if they had been specified by the user as arguments on the command line. </li>
```
magicDGS: The last sentence in this paragraph should be a caveat to solve in the future. Maybe it should be noted that it will change because it is a bug described in https://github.com/broadinstitute/barclay/issues/23

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -76,103 +110,127 @@
...
+     * descriptor wants this class to be included in from the list of plugins discovered dynamically.
      *
-     * In addition, implementations should recognize and reject any attempt to instantiate
-     * a second instance of a plugin that has the same simple class name as another plugin
-     * controlled by this descriptor (which can happen if they have different qualified names
-     * within the base package used by the descriptor) since the user has no way to disambiguate
-     * these on the command line).
+     * @return false if the plugin class shouldn't be used; otherwise true
+     */
+    public boolean includePluginClass(Class<?> c) { return true;}
```
magicDGS: I think that this is wrong, because to include a class for a plugin it should be at least of class `T`, which may be reached by bounding the parameter or check if it is assignable from the base class... Otherwise, every class might be included, even if it is not of class `T`.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -76,103 +110,127 @@
...
+     * instance of a plugin that has the same simple class name as another plugin controlled by this
+     * descriptor (which can happen if they have different qualified names within the base package
+     * used by the descriptor) since the user has no way to disambiguate these on the command line).
+     *
+     * @param pluginClass a plugin class discovered by the command line parser that
+     *                       was not rejected by {@link #includePluginClass}
      * @return the instantiated object that will be used by the command line parser
      * as an argument source
      * @throws IllegalAccessException
      * @throws InstantiationException
```
magicDGS: describe when this exceptions are thrown if possible

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -76,103 +110,127 @@
...
      */
-    public abstract Object getInstance(Class<?> pluggableClass)
+    public abstract T createInstanceForPlugin(Class<?> pluginClass)
```
magicDGS: Why the pluginClass isn't bounded? Unless it is a factory class, it should be at least `T`, no?

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -76,103 +110,127 @@
...
+     * arguments with any default plugins that have been provided to this descriptor. This list
+     * represents the plugins that will actually be used by the consumer. The resolution policy is
+     * descriptor-dependent.
+     *
+     * NOTE: calling this method before argument parsing (and thus before {@link #validateAndResolvePlugins}
+     * has been called) may return a different list than calling it after parsing, since the resolution
+     * policy is implementation-dependent and may depend on the actual arguments specified by the user
+     * on the command line.
      */
-    public abstract List<T> getAllInstances();
```
magicDGS: As the plugin specifies that all the instances should be kept in the descriptor, maybe it is a good idea to be able to have a getter for all the instances too (this will be useful for some purposes, like documentation).

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -76,103 +110,127 @@
...
+
+    /**
+     * Return the class object for the plugin with simple class name {@code pluginName}
+     * Used for help/usage and documentation generation.
+     *
      * @param pluginName Name of the plugin requested
      * @return Class object for the plugin instance requested
      */
-    public abstract Class<?> getClassForInstance(final String pluginName);
+    public abstract Class<?> getClassForPluginHelp(final String pluginName);
```
magicDGS: Why isn't it bounded? I still think that the methods should be bounded because the plugin descriptor should be used with classes of `T` without exception...

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
cmnbroad: I'm not sure what we would say about this here. If and when we fix the bug, I think the only change here would be that this last sentence would be removed.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
cmnbroad: In principle, we could add a lower bound (<? super T>) for this particular case, but for most of the rest of the cases, a lower bound makes the makes the method inaccessible to callers that don't have compile-time knowledge of T (like the command line parser), and an upper bound requires awkward constructs in the implementation. So, as I mentioned above, I'm advocating that we just leave all of the Class params unbounded to keep them consistent.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
cmnbroad: We can't use a type bound here, even a lower bound, since it would preclude the Barclay CLP or any code that doesn't have compile time knowledge of T from calling the method. We COULD use isAssignableFrom in the default implementation as a substitute, if thats what you're suggesting. It would never fail though, unless there was a bug in Barclay, which it would mask. It does seem more correct though.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
cmnbroad: See above comment.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
cmnbroad: What would the use case, and contract, for that be ? What would "all" instances mean ? The reason we changed the name was to try to remove the ambiguity around that. Now we have getDefaultInstances, and getResolvedInstances, which have clear(er) meaning. I'm not sure how that would be used for doc ?

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
cmnbroad: See above.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
cmnbroad: Changed default implementation to use isAssignableFrom.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
magicDGS: But I think that it is useful to say that it is a bug that will be solved, and not a feature done in purpose. Like that, API users will be aware that it may change in the future...

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
magicDGS: Ok, I understand the problem now. But at least we should enforce the behavior in the contract by adding a `It must return a base class of type {@link T}` to the javadoc...

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -76,103 +110,127 @@
...
+    public abstract Class<?> getPluginBaseClass();
 
     /**
-     * Return an instance of the specified pluggable class. The descriptor should
-     * instantiate or otherwise obtain (possibly by having been provided an instance
-     * through the descriptor's constructor) an instance of this plugin class.
-     * The descriptor should maintain a list of these instances so they can later
-     * be retrieved by {@link #getAllInstances}.
+     * Determine if a plugin class should be retained for this descriptor. Return true if the
+     * descriptor wants this class to be included in from the list of plugins discovered dynamically.
```
magicDGS: We should enforce to return `false` for classes that are not lower bounded by `T`, at least at the contract level (javadoc).

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
magicDGS: It will be nice to have a javadoc comment that the param is supposed to be of class `T`, and maybe enforce to throw a `CastClassException` if the `pluginClass` is not lower-bounded.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
magicDGS: Because the plugin is enforced to maintain the instances generated by `createInstanceForPlugin`, this will be a nice getter for retrieving all of them. Maybe instead of `getAllInstances`it should be named `getCreatedInstances`.

For documentation, it might be useful to get all the possible classes names in one shot, and check if all of them have their own documentation. An example, a custom doclet may use all the instances to retrieve the documentation of them in the plugin itself: this cannot be done with `getAllowedValuesForDescriptorHelp()`, because they are simple names.

Nevertheless, this is no high priority and we can open an issue to do it in the future...

magicDGS: O, I just forgot that `getClassForPluginHelp` is exactly doing that for the `String` name. So they can be retrieved by iterating over `getAllowedValuesForDescriptorHelp` and using `getClassForPluginHelp` (although it may be inefficient, but good enough)

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
magicDGS: Enforce in the contract that the returned value should be lower-bounded for `T` and that it is safe to cast it.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -2,70 +2,104 @@
...
+ * <p>
+ * Each plugin descriptor passed to the command line parser is queried to determine the class and
+ * package names to be searched for plugin classes controlled by that descriptor. The descriptor
+ * object itself is  added to the parser's list of argument sources, and for each matching plugin class
+ * discovered in the packages list defined the descriptor, the parser delegates back to the descriptor
+ * to obtain an instance of that plugin. Each of these plugin objects are then also added to the parsers's
+ * list of argument sources.
+ * </p>
+ * <p>
+ * Descriptor instances should have at least one @Argument field used to accumulate the plugins that
```
jamesemery: Here you should mention that the pluginDescirptor is responsible for maintaining a list of instances discovered through `createInstanceForPlugin` for later validation and use, it is mentioned in some of the method comments below.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -76,103 +110,127 @@
...
+     * the descriptor to validate and resolve the plugin arguments that have been specified, and to
+     * establish any state necessary to implement {@link #getResolvedInstances}.
      *
-     * It is the descriptor's job to contain an argument list which will be populated
-     * by the command line parser with the name of each plugin specified on the command line,
-     * and for each such plugin, to maintain a list of the corresponding instance. This
-     * method gives the descriptor a chance to reduce that list to include only those
-     * instances actually seen on the command line.
+     * It is the descriptor's job to maintain a list of all plugins discovered by the command
+     * line parser (signaled by calls to {@link #createInstanceForPlugin}), and an @Argument
```
jamesemery: These requirements should probably be bumped up to the class comment

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -76,103 +110,127 @@
...
+     * field to contain name of each plugin specified by the user on the command line. This
+     * method gives the descriptor a chance to validate and resolve those two lists against one
+     * another, and to derive a reduced list that contains only those instances that actually
+     * should be enabled, based on whatever resolution policy is in place for the descriptor.
```
jamesemery: Interestingly, the the implementation in GATKReadPluginDescriptor is actually doing none of this merging at this stage but rather at the get instance stage. Indeed this almost seems misleading as for any managed classes with arguments we can't have two instances floating around when we perform command line parsing.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -76,103 +110,127 @@
...
+     * represents the plugins that will actually be used by the consumer. The resolution policy is
+     * descriptor-dependent.
+     *
+     * NOTE: calling this method before argument parsing (and thus before {@link #validateAndResolvePlugins}
+     * has been called) may return a different list than calling it after parsing, since the resolution
+     * policy is implementation-dependent and may depend on the actual arguments specified by the user
+     * on the command line.
      */
-    public abstract List<T> getAllInstances();
+    public abstract List<T> getResolvedInstances();
```
jamesemery: GATKReadFilterDescriptor breaks this contract, it returns only those specified on the command line with no resolution between the default filters and user specified filters being performed.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
cmnbroad: Done.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -2,70 +2,104 @@
...
+ * </p>
+ * <p>
+ * Each plugin class controlled by a descriptor:
+ * </p>
+ * <ul>
+ * <li> Must live in one of the packages returned by the {@link #getPackageNames} method. </li>
+ * <li> Must have a type that is a subclass of the plugin base class returned by the {@link #getPluginBaseClass()}
+ *      method </li>
+ * <li> Must have a unique simple class name in order to avoid command line name collisions </li>
+ * <li> Should contain one or more @Argument annotated fields for any arguments used with that plugin. Plugin arguments
```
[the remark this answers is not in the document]
cmnbroad: Changed "should" to "may".

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
cmnbroad: Yeah, I think the language is too strong. The first line of this comment says "and to establish any state necessary to implement {@link #getResolvedInstances}", which I think is the essential thing. The language above, where I say "..and resolve those two lists..." is probably to prescriptive, so I'll relax that a bit. It doesn't matter whether the resolution is done here, or in getResolvedInstances; only that once this method is called, getResolvedInstances should have the state necessary to render a resolved list.

This PR is describing a change to the existing contract though, so once we assimilate it into GATK, we'll have to change the implementation of the descriptors to honor the new contract. I think the current GATK read filter descriptor does the "resolution" part of this in "getMergedReadFilter". To conform to this new contract, that will have to be moved into getResolvedInstances. I have a branch where I made those changes to test this all out. Once we've completed the review of this branch, I'll probably run it with that again to make sure it all works.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
cmnbroad: Right - this is describing a new contract, and that difference is the essential part of the change. "all" instances is ambiguous. So we're moving to getDefaultInstances and getResolvedInstances. Hopefully less ambiguous.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
```diff
@@ -2,70 +2,104 @@
...
+ * </ul>
+ * <li> When command line argument parsing is complete (all command line arguments have been processed), but
+ * before the parser returns control to the caller, a one time call: </li>
+ * <ul>
+ * <li> {@link #validateAndResolvePlugins} </li>
+ * </ul>
+ * <li> After the command line parser returns from argument parsing. These methods are generally called by
+ * the application that consumes the plugins: </li>
+ * <ul>
+ * <li> {@link #getDefaultInstances} - whenever a plugin consumer wants the list of plugin instances that were
```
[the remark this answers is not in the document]
cmnbroad: To me that seems more like a property of the resolution policy of the specific descriptor implementation, not something that should be prescribed by this contract. The cases we know about require that (read filters and annotations), because having duplicates make no sense. But I could imagine a plugin descriptor that would relax that. Let me know if you still think that

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
cmnbroad: Done above.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
cmnbroad: Not much to add there, but done.

src/main/java/org/broadinstitute/barclay/argparser/CommandLinePluginDescriptor.java
cmnbroad: Updated all the javadoc to include the type bounds where appropriate.

PR #41, opened 2017-02-10
Fix unbounded null throwing exception

### LINKED ISSUE #40: Bug: null numeric unbounded arguments try thrown IllegalArgumentException
When unbounded arguments have a null value, the command line parser try to thrown an `CommandLineException.OutOfRangeArgumentValue` but because there are no boundaries it throws an `IllegalArgumentException`.

magicDGS:
- Fixes #40 
- Add previously failing test

### DIFF
--- a/src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
+++ b/src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
@@ -535,7 +535,7 @@ public final class CommandLineArgumentParser implements CommandLineParser {
         final Double argumentDoubleValue = (argumentValue == null) ? null : ((Number)argumentValue).doubleValue();
 
         // Check hard limits first, if specified
-        if (isOutOfRange(argumentDefinition.minValue, argumentDefinition.maxValue, argumentDoubleValue)) {
+        if (argumentDefinition.hasBoundedRange() && isOutOfRange(argumentDefinition.minValue, argumentDefinition.maxValue, argumentDoubleValue)) {
             throw new CommandLineException.OutOfRangeArgumentValue(argumentDefinition.getLongName(), argumentDefinition.minValue, argumentDefinition.maxValue, argumentValue);
         }
         // Check recommended values
@@ -552,6 +552,7 @@ public final class CommandLineArgumentParser implements CommandLineParser {
                 logger.warn("Argument --{} has value {}, but maximum recommended is {}",
                         argumentDefinition.getLongName(), argumentDoubleValue, argumentDefinition.maxRecommendedValue);
             }
+            // if it does not have any recommended value, it does not log anything as expected
         }
     }
 
@@ -1184,6 +1185,11 @@ public final class CommandLineArgumentParser implements CommandLineParser {
 
         public boolean isFlag(){
             return field.getType().equals(boolean.class) || field.getType().equals(Boolean.class);
+        }
+
+        /** Returns {@code true} if the argument has a bounded range; {@code false} otherwise. */
+        public boolean hasBoundedRange() {
+            return this.minValue != Double.NEGATIVE_INFINITY || this.maxValue != Double.POSITIVE_INFINITY;
         }
 
         /**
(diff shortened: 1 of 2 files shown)


magicDGS:
Could you please have a look to this @cmnbroad?  It is a very simple fix for a bug that is preventing me to update my dependencies to the latest GATK/Barclay versions... Thank you in advance.

cmnbroad:
Hi @magicDGS Yes, will take a look tomorrow.

magicDGS:
Thank you very much @cmnbroad!

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -1186,6 +1187,11 @@ public boolean isFlag(){
             return field.getType().equals(boolean.class) || field.getType().equals(Boolean.class);
         }
 
+        /** Returns {@code true} if the argument has a bounded range; {@code false} otherwise. */
+        public boolean hasBoundedRange() {
+            return this.minValue != Double.NEGATIVE_INFINITY || this.maxValue != Double.POSITIVE_INFINITY;
+        }
+
```
cmnbroad: I think this could be private for now. Also, now that this method exists, we should use it in the other places where we do the same check (i.e., in the argdef constructor). Likewise, we should add a hasRecommendedRange and use that at the appropriate call sites.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
```diff
@@ -841,6 +841,9 @@ public void testWithDoubleBoundariesArgumentsForInteger() {
         // recommended values are not explicitly verified by the tests, but do force the code through the warning code paths
         @Argument(doc = "Integer in the range [0, 30]", optional = true, minValue = 0, minRecommendedValue = 10, maxRecommendedValue = 15, maxValue = 30)
         public int integerArg = 20;
+        // Integer without boundaries and null should be allowed
+        @Argument(doc = "Integer with null value allowed", optional = true)
+        public Integer nullInteger = null;
     }
```
cmnbroad: Its confusing to have this argument, and the corresponding test, live inside the "WithBoundariesArguments" class and test. It should live somewhere else, maybe in its own test if there is no current test thats appropriate. We should also add a similar test case for a Double (with null) arg.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -535,7 +535,7 @@ private void checkArgumentRange(final ArgumentDefinition argumentDefinition, fin
         final Double argumentDoubleValue = (argumentValue == null) ? null : ((Number)argumentValue).doubleValue();
 
         // Check hard limits first, if specified
-        if (isOutOfRange(argumentDefinition.minValue, argumentDefinition.maxValue, argumentDoubleValue)) {
+        if (argumentDefinition.hasBoundedRange() && isOutOfRange(argumentDefinition.minValue, argumentDefinition.maxValue, argumentDoubleValue)) {
```
cmnbroad: We should add a hasRecommendedRange method, similar to hasBoundedRange, and use that below where we check the recommended range.

magicDGS: It is not necessary for the logging path, because if there is no range there are no logging. Anyway, I included the check and added an exception (should never reach).

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -552,6 +552,7 @@ private void checkArgumentRange(final ArgumentDefinition argumentDefinition, fin
                 logger.warn("Argument --{} has value {}, but maximum recommended is {}",
                         argumentDefinition.getLongName(), argumentDoubleValue, argumentDefinition.maxRecommendedValue);
             }
+            // if it does not have any recommended value, it does not log anything as expected
```
magicDGS: I removed the comment and added an exception, because now it should not happen.

magicDGS: Sorry, this should be the comment. I misunderstood where it is....

magicDGS:
I addressed your comments in the last commit. Thank you very much @cmnbroad -- back to you

magicDGS:
I think that this is ready to go, @cmnbroad. It is important for me to update dependencies with the new changes in some of my projects. Thank you in advance!

magicDGS:
Thanks for this!

PR #42, opened 2017-02-13
Use field name in doc when annotation has no arg name.

cmnbroad:
The current doc output contains no argument name if the Argument annotation doesn't specify one, but it should use the field name.

Also fixes the test for JSON doc output, which was redundantly checking the html output file instead of the JSON output file.


droazen:
:+1: looks good, merge when ready

PR #44, opened 2017-02-27
Add tests for bounded collection args.

cmnbroad:
Just noticed we don't have any tests for collections with ranges, but we should.

### DIFF
--- a/src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
+++ b/src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
@@ -939,6 +939,56 @@ public final class CommandLineArgumentParserTest {
         return result;
     }
 
+    @CommandLineProgramProperties(
+            summary = "tool with bounded collection argument",
+            oneLineSummary = "bounded collection argumen",
+            programGroup = TestProgramGroup.class
+    )
+    public class BoundedCollection {
+        @Argument(doc = "Integer collection in the range [0, 30]",
+                optional = true,
+                minValue = 0, maxValue = 30,
+                minRecommendedValue = 10, maxRecommendedValue = 15)
+        public List<Integer> intListArg = new ArrayList<>();
+    }
+
+    @DataProvider(name="goodBoundedCollectionArgs")
+    public Object[][] goodBoundedCollection() {
+        return new Object[][]{
+                {new String[]{"--intListArg", "1"}, new Integer[]{ 1 } },
+                {new String[]{"--intListArg", "19", "--intListArg", "2"}, new Integer[]{ 19, 2 } },
+                {new String[]{"--intListArg", "0", "--intListArg", "18", "--intListArg", "27" }, new Integer[]{ 0, 18, 27 } },
+        };
+    }
+
+    @Test(dataProvider = "goodBoundedCollectionArgs")
+    public void testGoodBoundedCollections(final String[] argv, final Integer[] expectedIntegers) throws Exception {
+        final BoundedCollection o = new BoundedCollection();
+        final CommandLineArgumentParser clp = new CommandLineArgumentParser(o);
+        Assert.assertTrue(clp.parseArguments(System.err, argv));
+        Assert.assertEquals(expectedIntegers.length, o.intListArg.size());
+        for (int i = 0; i < expectedIntegers.length; i++) {
+            Assert.assertEquals(o.intListArg.get(i).intValue(), expectedIntegers[i].intValue());
+        }
+    }
+
+    @DataProvider(name="badBoundedCollectionArgs")
+    public Object[][] badBoundedCollection() {
+        return new Object[][]{
+                {new String[]{"--intListArg", "-1"}},
+                {new String[]{"--intListArg", "57"}},
+                {new String[]{"--intListArg", "1", "--intListArg", "57"}},
+                {new String[]{"--intListArg", "null" }},
+        };
+    }
+
+    @Test(dataProvider = "badBoundedCollectionArgs", expectedExceptions = CommandLineException.OutOfRangeArgumentValue.class)
+    public void testBadBoundedCollections(final String[] argv) throws Exception {
+        final BoundedCollection o = new BoundedCollection();
+        final CommandLineArgumentParser clp = new CommandLineArgumentParser(o);
+        Assert.assertTrue(clp.parseArguments(System.err, argv));
+    }
+
     @Test(expectedExceptions = CommandLineException.CommandLineParserInternalException.class)
     public void testHiddenRequiredArgumentThrowException() throws Exception {
         @CommandLineProgramProperties(summary = "tool with required and hidden argument",


src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
```diff
@@ -939,6 +939,56 @@ public void testOutOfRangesArguments(final String[] argv) throws Exception {
...
+        }
+    }
+
+    @DataProvider(name="badBoundedCollectionArgs")
+    public Object[][] badBoundedCollection() {
+        return new Object[][]{
+                {new String[]{"--intListArg", "-1"}},
+                {new String[]{"--intListArg", "57"}},
+                {new String[]{"--intListArg", "1", "--intListArg", "57"}},
+                {new String[]{"--intListArg", "null" }},
```
magicDGS: Is there any test for this if the collection does not have bounded arguments? Maybe this corresponds here too.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
```diff
@@ -939,6 +939,56 @@ public void testOutOfRangesArguments(final String[] argv) throws Exception {
...
+    public Object[][] goodBoundedCollection() {
+        return new Object[][]{
+                {new String[]{"--intListArg", "1"}, new Integer[]{ 1 } },
+                {new String[]{"--intListArg", "19", "--intListArg", "2"}, new Integer[]{ 19, 2 } },
+                {new String[]{"--intListArg", "0", "--intListArg", "18", "--intListArg", "27" }, new Integer[]{ 0, 18, 27 } },
+        };
+    }
+
+    @Test(dataProvider = "goodBoundedCollectionArgs")
+    public void testGoodBoundedCollections(final String[] argv, final Integer[] expectedIntegers) throws Exception {
```
magicDGS: Why `expectedIntegers` is not `int[]`?

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
```diff
@@ -939,6 +939,56 @@ public void testOutOfRangesArguments(final String[] argv) throws Exception {
...
+            summary = "tool with bounded collection argument",
+            oneLineSummary = "bounded collection argumen",
+            programGroup = TestProgramGroup.class
+    )
+    public class BoundedCollection {
+        @Argument(doc = "Integer collection in the range [0, 30]",
+                optional = true,
+                minValue = 0, maxValue = 30,
+                minRecommendedValue = 10, maxRecommendedValue = 15)
+        public List<Integer> intListArg = new ArrayList<>();
```
magicDGS: What will happen if the default list includes a value out of range? It will blow up? Maybe that also requires a test...

magicDGS: And what will happen if there are default values and `--intListArg null --intListArg 15` to clean the defaults?

cmnbroad: For the first one, there is neither tests nor code to check that the initial value is within range. I added code that throws an internal exception if the initial value is out of range, for both scalar and collection, and tests.

The second one is already illegal. Since the default behavior of this parser for collections is "replace" rather than "append", the parser already enforces that if you pass "null" as a value for a collection, you cannot pass any other values for that arg.  If you want to reset the list to a single value of 15, then you just say "--intListArg 15." Finally, I believe we already have a test to validate that "null" is not a valid value for something with a range (since null is always outside the range); I added an additional test to verify that for collections as well.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
cmnbroad: Only because Integer is the type of the value in the argument, so it makes the Assert overload unambiguous. I did remove the unnecessary .getInt calls from the Asserts though.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
cmnbroad: Yes, see testClearDefaultValuesFromListArgumentAndAddNew.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
magicDGS: Oh, I forgot that now is replace instead of append. Thanks for remind me.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
magicDGS: Ok, I understand it now.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
magicDGS: And thanks for the check of the initial value out of the range.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
magicDGS: So it is illegal to provide `null` always? I mean, if the intListArg in this case does not have boundaries, is it forbiden to provide `--intListArg null` (without any other argument)? And the same case for any other object collection....

cmnbroad: No. If there is no range specified, its perfectly legal to use null to reset the collection (as long as there are no other arguments). But I think you're right that there is no positive test for this, only the negative test to verify that you can't provide other values along with null. I'll add one. At some point I really want to refactor these tests since its really hard to tell whats covered and whats not.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
```diff
@@ -534,6 +534,22 @@ public void testCollectionThatCannotBeAutoInitialized() {
 
     }
 
+    // Should be able to use null to reset a list (with no ranges specified) to empty
+    @Test
+    public void testResetListToNull() {
+        class ResetListToNull {
+            @Argument
+            public List<String> LIST = makeList("foo", "bar");
```
magicDGS: I will use instead a numeric list to check that this is correct. That was one of my bugs (see #41) and maybe it is better to be sure that it is not broken also for collections.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -1212,6 +1165,80 @@ private boolean hasRecommendedRange() {
...
+                            String.format("Min/max value ranges can only be set for numeric arguments. "+
+                                    "Argument --%s has a minimum or maximum value but has a non-numeric type.",
+                                    this.getLongName()));
+                }
+                if (Integer.class.isAssignableFrom(this.type)) {
+                    if (!isInfinityOrMathematicalInteger(this.maxValue)
+                            || !isInfinityOrMathematicalInteger(this.minValue)
+                            || !isInfinityOrMathematicalInteger(this.maxRecommendedValue)
+                            || !isInfinityOrMathematicalInteger(this.minRecommendedValue)) {
+                        throw new CommandLineException.CommandLineParserInternalException(
```
LeeTL1220: Can the maximum (or minimum) value be infinity? If so, this will fail.  I agree that users should not specify these values, but perhaps this error message should include the suggestion not to specify infinity.

src/test/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParserTest.java
```diff
@@ -875,6 +892,69 @@ public void testWithDoubleBoundariesArgumentsForInteger() {
...
+        }
+        final BoundedIntegerIntializedOutsideRange o = new BoundedIntegerIntializedOutsideRange();
+        final CommandLineArgumentParser clp = new CommandLineArgumentParser(o);
+    }
+
+    @Test(expectedExceptions = CommandLineException.CommandLineParserInternalException.class)
+    public void testBoundedCollectionIntializedOutOfRange() throws Exception {
+        @CommandLineProgramProperties(
+                summary = "tool with bounded collection argument with bad initial value",
+                oneLineSummary = "ool with bounded collection argument with bad initial value",
```
LeeTL1220: Typo.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -1212,6 +1165,80 @@ private boolean hasRecommendedRange() {
...
+            final Double argumentDoubleValue = (argumentValue == null) ? null : ((Number)argumentValue).doubleValue();
+
+            // Check hard limits first, if specified
+            if (hasBoundedRange() && isOutOfRange(minValue, maxValue, argumentDoubleValue)) {
+                throw new CommandLineException.OutOfRangeArgumentValue(getLongName(), minValue, maxValue, argumentValue);
+            }
+            // Check recommended values
+            if (hasRecommendedRange() && isOutOfRange(minRecommendedValue, maxRecommendedValue, argumentDoubleValue)) {
+                final boolean outMinValue = minRecommendedValue != Double.NEGATIVE_INFINITY;
+                final boolean outMaxValue = maxRecommendedValue != Double.POSITIVE_INFINITY;
```
LeeTL1220: Do we need to check for NaN?  This may be a matter of opinion...

PR #51, opened 2017-05-03
Add BetaFeature marker interface and integrate with command line u…

cmnbroad:
…sage and doc gen.


droazen:
As discussed in person, rename to `BetaFeature`, as "experimental" is actually too strong a term for this purpose. Go ahead and merge after the rename.

Issue #52, opened 2017-05-05
mutex behaves inconsistently with collection arguments and optional=false


lbergelson:
Mutex arguments that aren't collections allow both arguments to be marked as `optional=false`.  This is treated as "exactly of these arguments is required"  However collection arguments don't behave the same way.  A scalar and collection argument where both are mutually exclusive to each other and both are `optional=false` will fail if the collection argument is not specified.

PR #53, opened 2017-05-10
Fix mutex for mixed scalar/collection arguments.

cmnbroad:
Fixes https://github.com/broadinstitute/barclay/issues/52.


src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -462,7 +462,7 @@ private void assertArgumentsAreValid()  {
                 if (argumentDefinition.isCollection && !argumentDefinition.optional) {
                     @SuppressWarnings("rawtypes")
                     final Collection c = (Collection) argumentDefinition.getFieldValue();
-                    if (c.isEmpty()) {
+                    if (c.isEmpty() && mutextArgumentNames.length() == 0) {
                         throw new CommandLineException.MissingArgument(fullName, "Argument '" + fullName + "' must be specified at least once.");
```
lbergelson: Could you update the missing argument message so it includes the "unless any of X" were specified clause that the message below has?

lbergelson:
@cmnbroad Looks good, one minor request and then free to merge.

Issue #39, opened 2017-02-09
Design index template(s)


vdauwera:
I'm starting to plan how we'll make the tooldocs available on the GATK website, and specifically what organization layouts we want to offer to users. There are three main patterns that people tend to follow when looking for docs, which would be best served by offering separate index pages:

- full alphabetical list of all tools
- subset by top-level package (in the GATK world, core vs protected vs Picard), then alphabetical
- functional breakdown (QC vs bam processing vs variant discovery etc)

Note that this would only apply to tools, proper -- read filters, annotation modules, metrics collections (in Picard) etc still make sense to categorize separately in any case. 

That being said, I need to think a bit more about the UX side of things before implementing anything. TBC. Comments welcome.

vdauwera:
Yo @droazen @lbergelson can you make me a writer on this repo?

cmnbroad:
@vdauwera I added you as a writer. One comment - depending on how your thinking evolves, it may be better to keep some of what you're describing in the GATK or Picard repositories, so Barclay can remain agnostic about who the docgen consumer is.

vdauwera:
Thanks! 

I agree about keeping gatk/Picard specific details out of barclay; but I'm thinking it may be useful to make index templates that provide generic schemas for the different patterns of organization. So I plan to work from here and extract things that become too specific.

PR #58, opened 2017-05-23
Implement static start method for HelpDoclet

### LINKED ISSUE #55: HelpDoclet cannot be used directly
Although the javadoc states that  `this class can be used to generate documentation directly`, that's not the case because if used in a gradle task (e.g. onlineDoc) it complains with the following error:

```
:compileJava UP-TO-DATE
:processResources UP-TO-DATE
:classes UP-TO-DATE
:cleanOnlineDoc
:onlineDoc
javadoc: error - Doclet class org.broadinstitute.barclay.help.HelpDoclet does not contain a start method
1 error
:onlineDoc FAILED
```

It seems that it is because it requires a public static method called start, as in the GATK Doctlet or in the Picard one. I'm planing to have my own doctlet, but while testing how this work it will be nice to be able to use the minimal implementation...

magicDGS:
This is required to allow the usage of the HelpDoclet without extension
of the class. Includes an update of the javadoc to clarify that the
method is required.

Close #55


magicDGS:
Small bug and documentation fix for make `HelpDoclet` usable, @cmnbroad. Can you review, please?

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
```diff
@@ -82,6 +84,14 @@
     private Set<DocWorkUnit> workUnits;     // Set of all things we are going to document
 
     /**
+     * The entry point to javadoc generation. It calls {@link #startProcessDocs(RootDoc)} (String[])} from this instance.
+     * Consumers should implement this static method to call {@link #startProcessDocs(RootDoc)} from their instance.
```
cmnbroad: Since this method is static, it probably shouldn't mention "this instance". Can you reword this to something like "The entry point for javadoc generation. It creates an instance of HelpDoclet and calls startProcessDocs on that instance. Custom Barclay doclets should subclass this class, and implement a similar static method that creates an instance of the doclet subclass and delegates to that instance's startProcessDocs".

magicDGS: Done with small modifications.

magicDGS:
I will add a test, but for that I will also need to modify the help templates and copy the current ones to the test resources for the test doclet. Do you mind if I do this, @cmnbroad? The custom templates are not working because they mention the 'supercat' tag and maybe others (I haven't look into detail). I do not want to go too much into this (better to design this while addressing #39), but having non-specific template will help with development of custom template.

My idea behind these changes is to have working the custom doclet to make easy to have a quick check if the classes are working with the default one and the default templates. Then, if the user need it, they could add custom features and make it more complicated.

Waiting for your answer here, @cmnbroad!

cmnbroad:
@magicDGS Yeah, just a simple test using simple templates should do. It will ensure that we don't break this use case. Preferably put them in a separate folder (you might want to move the existing ones so the two sets can live in parallel folders that correspond to the tests).

magicDGS:
I'll do it.

My idea was more related to change the structure of resources by the following:

* Move the current templates to the `test/resources/testTemplates` as they are
* Derive new default templates from them, simplifying specific-features tested in the current implementation.
* These templates could be used as started point for custom templates (as it is stated in the current Wiki). They will leave in `main/resources/org/broadinstitute/barclay/helpTemplates`, allowing to pack them into the jar file and use them as-is in projects that requires a very simple documentation template.

What do you think about it, @cmnbroad? I can do this here or in a different ticket, but I believe that this organization may benefit other user with less requirements....

cmnbroad:
@magicDGS That sounds like a pretty good plan. Thanks.

magicDGS:
I repackaged the templates, removed in the supercat in the default ones, and added a test for the default doclet.

Thanks for reviewing - back to you, @cmnbroad!

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
```diff
@@ -41,6 +41,8 @@
  * {@link #createWorkUnit}
  * {@link #createGSONWorkUnit}
  *
+ * This requires that the static {@link #start(RootDoc)} methods is implemented to call the
+ * {@link #startProcessDocs(RootDoc)} from their instance.
```
cmnbroad: I would remove the blank line (43) so this appears with the list right above, and change it to say:

{@link #start(RootDoc)} A static method that instantiates the subclass and delegates to the instance method {@link #startProcessDocs(RootDoc)}.

src/test/java/org/broadinstitute/barclay/help/DocumentationGenerationIntegrationTest.java
```diff
@@ -65,4 +65,44 @@ public boolean filesContentsIdentical(
...
+
+        outputDir.deleteOnExit();
+
+        String[] argArray = new String[]{
+                "-build-timestamp", "2016/01/01 01:01:01",      // dummy, constant timestamp
+                "-absolute-version", "11.1",                    // dummy version
+                "-settings-dir", "src/main/resources/org/broadinstitute/barclay/helpTemplates/",
+                "-d", outputDir.getAbsolutePath(),
+                "-output-file-extension", "html",
+                "-doclet", TestDoclet.class.getName(),
```
magicDGS: A rebase problem, sorry. I change it!

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java

src/test/java/org/broadinstitute/barclay/help/DocumentationGenerationIntegrationTest.java
```diff
@@ -65,4 +65,44 @@ public boolean filesContentsIdentical(
         return Arrays.equals(actualBytes, expectedBytes);
     }
 
+    @Test
+    public void testDefaultDoclet() throws IOException {
```
magicDGS: Will do in second commit.

magicDGS: Done.

magicDGS:
I addressed your comments. Back to you @cmnbroad!

cmnbroad:
@magicDGS Looks good - thanks for doing all that!

magicDGS:
Thanks!

PR #60, opened 2017-05-24
Index output extension option

### LINKED ISSUE #59: Allow a different output extension for the index file
I'm in the case in which I would like to have a different format for the index than the rest of components for use with jekyll: 1) index as an yml data file for iterate over different components,  and 2) markdown to render for each component.

It will be nice to have a new option for the index extension, which by default should be the same as the component extension. Someone have any objection to this?

magicDGS:
Includes the following changes:

- Make public some getters and implement new ones. This will be helpful for DocWorkUnitHandler implementations.
- Use getters within the class to use overriden ones from sub-classes.
- New option and field (including getter) for different index extension, which default outputFileExtension.

Close #59

### DIFF
--- a/src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
+++ b/src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
@@ -219,12 +219,17 @@ public class HelpDoclet {
     /**
      * @return the output extension to use, i.e., ".html" or ".php"
      */
-    protected String getOutputFileExtension() { return outputFileExtension; }
+    public String getOutputFileExtension() { return outputFileExtension; }
 
     /**
      * @return the name of the index template to be used for this doclet
      */
-    protected String getIndexTemplateName() { return "generic.index.template.html"; }
+    public String getIndexTemplateName() { return "generic.index.template.html"; }
+
+    /**
+     * @return the file where the files will be output
+     */
+    public File getDestinationDir() { return  destinationDir; }
 
     /**
      * Determine if a particular class should be included in the output. This is called by the doclet
@@ -332,7 +337,7 @@ public class HelpDoclet {
    ) throws IOException {
         // Get or create a template and merge in the data
         final Template template = cfg.getTemplate(getIndexTemplateName());
-        final File indexFile = new File(destinationDir + "/index." + outputFileExtension);
+        final File indexFile = new File(getDestinationDir() + "/index." + outputFileExtension);
         try (final FileOutputStream fileOutStream = new FileOutputStream(indexFile);
              final OutputStreamWriter outWriter = new OutputStreamWriter(fileOutStream)) {
             template.process(groupIndexMap(workUnitList, groupMaps), outWriter);
@@ -364,8 +369,8 @@ public class HelpDoclet {
 
         root.put("data", data);
         root.put("groups", groupMaps);
-        root.put("timestamp", buildTimestamp);
-        root.put("version", absoluteVersion);
+        root.put("timestamp", getBuildTimeStamp());
+        root.put("version", getBuildVersion());
 
         return root;
     }
@@ -441,7 +446,7 @@ public class HelpDoclet {
         try {
             // Merge data-model with template
             Template template = cfg.getTemplate(workUnit.getTemplateName());
-            File outputPath = new File(destinationDir + "/" + workUnit.getTargetFileName());
+            File outputPath = new File(getDestinationDir() + "/" + workUnit.getTargetFileName());
             try (final Writer out = new OutputStreamWriter(new FileOutputStream(outputPath))) {
                 template.process(workUnit.getRootMap(), out);
             }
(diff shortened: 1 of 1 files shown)


magicDGS:
This will be useful for my work using the Barclay doclet to document my tools. Thanks a lot @cmnbroad!

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
```diff
@@ -74,6 +75,7 @@
     protected static File settingsDir = DEFAULT_SETTINGS_DIR;
     protected static File destinationDir = DEFAULT_DESTINATION_DIR;
     protected static String outputFileExtension = DEFAULT_OUTPUT_FILE_EXTENSION;
+    protected static String indexFileExtension = null;
```

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
```diff
@@ -103,6 +105,14 @@ protected boolean startProcessDocs(final RootDoc rootDoc) throws IOException {
...
                 outputFileExtension = options[1];
             }
+            if (options[0].equals(INDEX_FILE_EXTENSION_OPTION)) {
+                indexFileExtension = options[1];
+            }
+        }
+
+        // default to outputFileExtension
+        if (indexFileExtension == null) {
+            indexFileExtension = outputFileExtension;
```
magicDGS: I did it like that to assing the value of `-output-file-extension` to the indexFileExtension in case that the `-index-file-extension` option is not provided. I think that this is better than provide the two of them in the case that they should be the same.

I kept this and assign `indexFileExtension` to null before parsing options to maintain this behavior. If you want me to change it, just let me know...

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
```diff
@@ -219,12 +229,22 @@ private void processDocs(final RootDoc rootDoc) {
     /**
      * @return the output extension to use, i.e., ".html" or ".php"
      */
-    protected String getOutputFileExtension() { return outputFileExtension; }
+    public String getOutputFileExtension() { return outputFileExtension; }
+
+    /**
+     * @return the output extension to use for the index, i.e., ".html" or ".php"
+     */
+    public String getIndexFileExtension() { return (indexFileExtension == null) ? outputFileExtension : indexFileExtension; }
```

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
```diff
@@ -89,6 +91,9 @@
      * @throws java.io.IOException if output can't be written.
      */
     protected boolean startProcessDocs(final RootDoc rootDoc) throws IOException {
+        // set the indexFileExtension to null to identify if it was provided with -index-file-extension option
+        // this is necessary the make the default the same as the -output-file-extension if it wasn't provided
+        indexFileExtension = null;
```
magicDGS: Ok, I will keep them independent. I wanted to make it flexible and allow the user to only specify one of them for consistency unless specified. For example: everything in Markdown, unless you specify that the index should be an yml file.

Anyway, I removed this in the next commit.

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
```diff
@@ -103,6 +108,14 @@ protected boolean startProcessDocs(final RootDoc rootDoc) throws IOException {
...
                 outputFileExtension = options[1];
             }
+            if (options[0].equals(INDEX_FILE_EXTENSION_OPTION)) {
+                indexFileExtension = options[1];
+            }
+        }
+
+        // default to outputFileExtension if it was not provided
+        if (indexFileExtension == null) {
+            indexFileExtension = outputFileExtension;
```
magicDGS: Done (the same as before).

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
```diff
@@ -219,12 +232,22 @@ private void processDocs(final RootDoc rootDoc) {
     /**
      * @return the output extension to use, i.e., ".html" or ".php"
      */
-    protected String getOutputFileExtension() { return outputFileExtension; }
+    public String getOutputFileExtension() { return outputFileExtension; }
```
magicDGS: I would like to have access to this paramater in my custom `DocWorkUnitHandler.getDestinationFilename()`, but with the previous implementation it is impossible unless the handler is in the same package as the Barclay one.

My use case is to generate a simpler name without the package and the same extension as the one provided in the command line. I also make public other getters because they may be useful for custom implementations: e.g., getIndexTemplateName will be useful to generate different indexes with the same template.

I will maintain this if you don't mind :)

magicDGS:
I did not implement the test yet, because that could be included after the test refactoring in #58 as a new param. There is also no test for the other output file extension, so maybe it could be do in a later PR to add test coverage to this methods?

Back to you @cmnbroad and thanks for reviewing!

magicDGS:
Thanks, I will open a PR to have a test with different index/output extension.

Issue #63, opened 2017-05-28
Docgen assumes DocumentedFeatures have no-arg constructors


cmnbroad:
Most documented features are command line programs, which have no-arg constructors, but some are not (i.e., TableReader), and so may not have a no-arg constructor. We need to relax the assumption and only instantiate classes that are command line programs.

Issue #62, opened 2017-05-28
Failure accessing private fields with javadoc.


cmnbroad:
When traversing through class hierarchies trying to resolve arguments, docgen checks each field ti encounters to see if its annotated as an argument collection, and if so, recursively reflects on the field's type. If the type has a private field that has javadoc, getDeclaredFields doesn't return the field, but javadoc includes it in the list of FieldDocs. The code needs to be tolerant of that case during field traversal. See https://github.com/broadinstitute/gatk-protected/issues/1048.

cmnbroad:
Fixed by #64.

PR #64, opened 2017-05-28
Don't require no-arg constructors for DocumentedFeatures.

cmnbroad:
Fixes https://github.com/broadinstitute/barclay/issues/63 and https://github.com/broadinstitute/barclay/issues/62 (originally discovered as https://github.com/broadinstitute/gatk-protected/issues/1048). Also removes the ridiculous test case that has an "extraDocs" reference to itself.

### DIFF
--- a/src/main/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandler.java
+++ b/src/main/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandler.java
@@ -180,21 +180,31 @@ public class DefaultDocWorkUnitHandler extends DocWorkUnitHandler {
             final List<Map<String, String>> featureMaps,
             final List<Map<String, String>> groupMaps) {
 
-        CommandLineArgumentParser clp;
+        CommandLineArgumentParser clp = null;
         List<? extends CommandLinePluginDescriptor<?>> pluginDescriptors = new ArrayList<>();
 
-        try {
-            final Object argumentContainer = workUnit.getClazz().newInstance();
-            if (argumentContainer instanceof CommandLinePluginProvider ) {
-                pluginDescriptors = ((CommandLinePluginProvider) argumentContainer).getPluginDescriptors();
-                clp = new CommandLineArgumentParser(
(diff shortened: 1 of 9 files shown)


src/test/java/org/broadinstitute/barclay/help/TestExtraDocs.java
```diff
@@ -5,7 +5,7 @@
 /**
  * Class for testing extraDocs property in docgen.
  */
-@DocumentedFeature(groupName = "Test extra docs group name", extraDocs = TestExtraDocs.class)
+@DocumentedFeature(groupName = "Test extra docs group name")
```
cmnbroad: extraDocs is still tested; its referenced in, and tested by, TestArgumentContainer. I removed it from here because its self-referential, which is harmless but unnecessary.

src/main/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandler.java
```diff
@@ -602,11 +614,16 @@ private FieldDoc getFieldDoc(final ClassDoc classDoc, final String name, final b
...
                 return fieldDoc;
 
+            // this can return null if the fieldDoc represents a field that has javadoc, but is private.
             final Field field = DocletUtils.getFieldForFieldDoc(fieldDoc);
-            if (field == null)
-                throw new RuntimeException("Could not find the field corresponding to " + fieldDoc +
-                        ", presumably because the field is inaccessible");
-            if (field.isAnnotationPresent(ArgumentCollection.class)) {
+            if (field == null) {
+                logger.warn(
```
cmnbroad: This code path is already trying to determine if the field is an ArgumentCollection, but it can't do that if it can't retrieve the Field. The problem with throwing here is that it aborts the whole run, even when it would succeed if it just kept searching. We only want to throw when we've exhausted the search and can't find the fieldDoc for an argument at all, which happens later on in the process.

magicDGS: I see, thanks for the explanation!

PR #68, opened 2017-06-02
Markdown integration

magicDGS:
Markdown integration in the CLI help and documentation generation using [commonmark-java](https://github.com/atlassian/commonmark-java). Includes:

- CLI: rendering to plain text fields in argparser annotations
- DocGen: rendering to HTML fields in argparser annotations and javadoc

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -48,6 +48,7 @@ dependencies {
     compile 'org.apache.commons:commons-lang3:3.4'
     compile 'org.apache.logging.log4j:log4j-api:2.3'
     compile 'org.apache.logging.log4j:log4j-core:2.3'
+    compile 'com.atlassian.commonmark:commonmark:0.9.0'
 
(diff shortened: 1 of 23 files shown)


magicDGS:
Here you have the PR, @vdauwera!

cmnbroad:
I haven't had a chance to look at this in detail yet (I'll get to it as soon as I can, but realistically it won't be until we get beta finished). One high-level request is that we find a way to incrementally unit-test new functionality like this at a granular level. We currently only have coarse-grained, file-based integration tests, and I'd like to find a way to avoid proliferating a new set of test files with each new feature, in addition to changing many/all of the existing ones. @magicDGS any thoughts on how we can address that ?

magicDGS:
I open an issue (#71) to discuss how to change the docgen tests. Let's keep this PR for the Markdown support, which may be also generate some discussion...

Issue #69, opened 2017-06-06
Allow usage of other tags and not only inline in DefaultDocWorkUnitHandler


magicDGS:
As an example:

```java
/**
* This is the javadoc for the description.
* 
* {@MyTag.test1 this is test1 tag}
*
* @MyTag.test2 this is test2 tag
*/
@DocumentedFeature
public class TestClass {}
```

If the custom tag prefix is `MyTag`, the final json will pass the `test1` tag but not the `test2` tag because `DefaultDocWorkUnitHandler.addCustomBindings` only uses the `ClassDoc.inlineTags()`. Using `ClassDoc.tags()` instead will allow to output `test2`, with the advantage of showing it in a normal javadoc task.

magicDGS:
Related to this, the method `DefaultDocWorkUnitHandler.addCustomBindings` could be overriden, but it could not access the  private method`getTagPrefix()`. It will be useful to make public (and final) that method to allow sub-classes to retrieve the prefix and perform filter in other javadoc-related documentation.

PR #70, opened 2017-06-06
Support all kind of tags in DefaultDocWorkUnitHandler

magicDGS:
Implementation of #69

### DIFF
--- a/src/main/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandler.java
+++ b/src/main/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandler.java
@@ -329,7 +329,8 @@ public class DefaultDocWorkUnitHandler extends DocWorkUnitHandler {
         }
     }
 
-    private String getTagPrefix() {
+    /** Gets the tag prefix as formatted in the javadoc; {@code null} if there is no user-defined prefix. */
+    public final  String getTagPrefix() {
         String customPrefix = getTagFilterPrefix();
         return customPrefix == null ?
                 customPrefix : "@" + customPrefix + ".";
--- a/src/main/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandler.java
+++ b/src/main/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandler.java
@@ -243,7 +243,7 @@ public class DefaultDocWorkUnitHandler extends DocWorkUnitHandler {
      */
     protected void addCustomBindings(final DocWorkUnit currentWorkUnit) {
         final String tagFilterPrefix = getTagPrefix();
-        Arrays.stream(currentWorkUnit.getClassDoc().inlineTags())
+        Arrays.stream(currentWorkUnit.getClassDoc().tags())
                 .filter(t -> t.name().startsWith(tagFilterPrefix))
                 .forEach(t -> currentWorkUnit.setProperty(t.name().substring(tagFilterPrefix.length()), t.text()));
     }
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandler.java
```diff
@@ -243,7 +243,7 @@ protected void addHighLevelBindings(final DocWorkUnit workUnit)
      */
     protected void addCustomBindings(final DocWorkUnit currentWorkUnit) {
         final String tagFilterPrefix = getTagPrefix();
-        Arrays.stream(currentWorkUnit.getClassDoc().inlineTags())
+        Arrays.stream(currentWorkUnit.getClassDoc().tags())
```
magicDGS: My aim here is to have tags similar to the javadoc `@see`, but customized. This will allow to remove from the help text this tags and even include them in the normal javadoc. For example:

```java
/**
 * This is the description of the documentation.
 * @barclay.type simple
 * @author magicDGS
 * @see OtherClass
 */
@DocumentedFeature
public class MyDocumentedClass {}
```

Maybe my change is not working (which is wierd, because it mimics a hack that I did in a custom project to include this feature), but my aim is to populate the "barclay.type" in the JSON, but ignoring it in the documentation text description. In Barclay, the test for round trip is testing the JSON, but the tags are not populated there anyway (I haven't change the expected files); we should either have a test JSON object or implement mocking tests for all this functionality.

I also notice another problem with the tags: if there are repeated tags, they are overridden. For instance, I have the situation when I want to add custom tags with "notes", "warnings", "caveats" and then populate them into the JSON to format them as a list. That's actually why I started with this: I do not want to have them embedded in the description (in the format `{@doc.warning be careful}`), but being able to process them with the `@doc.warning be careful` format to remove them from the doc test. Nevertheless, the problem is when two warnings should be separated, and then one overrides the other.

Maybe we should discuss about the tag-support and how it should look like in a different issue, and I should close this PR til we agree on the final shape of this functionality...

cmnbroad: @magicDGS I think your example above will work as long as there is no other text in between these two lines:

```
 * @barclay.type simple
 * @author magicDGS
```
But I think it would break if it looked like this:

```
 * @barclay.type simple
 * Other comment included in between tags
 * @author magicDGS
```

In that case, the custom @barclay.type tag's text will include everything up to  the start of the next comment: "simple * Other comment included in between tags".

cmnbroad: Also, I realized from looking into this more, and from looking at the GATK beta doc output, that there is a bug in Barclay docgen where in some cases we don't include javadoc after the first embedded tag, at least in some cases. I'll try to fix that ASAP since its fairly high priority. In the meantime, lets leave this PR open for the short term so we can retain this discussion thread, even though we'll probably replace it with another PR with a more comprehensive fix.

magicDGS: Actually that's a normal javadoc processing feature. These kind of tags (`@see`, `@author`, etc.) are suppose to be at the end of the javadoc instead of embedded in the text. That's why embedded tags such as `@link` have the curly-brackets syntax.

I think that the fix is to point out that this is like that... and actually IntelliJ re-format the javadoc putting them at the end, so I think that it's fair enough to keep it as it is now...

PR #73, opened 2017-06-12
Upgrade FreeMarker to 2.3.23

magicDGS:
I would like to upgrade FreeMarker because the new version comes with more built-ins for different kind of variables.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -55,7 +55,7 @@ dependencies {
     final javadocJDKFiles = files(((URLClassLoader) ToolProvider.getSystemToolClassLoader()).getURLs())
     compileOnly(javadocJDKFiles)
     testCompile(javadocJDKFiles)
-    compile 'org.freemarker:freemarker:2.3.18'
+    compile 'org.freemarker:freemarker:2.3.23'
     compile 'com.google.code.gson:gson:2.2.2'
 
     testCompile 'org.testng:testng:6.9.6'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -55,7 +55,7 @@ dependencies {
     final javadocJDKFiles = files(((URLClassLoader) ToolProvider.getSystemToolClassLoader()).getURLs())
     compileOnly(javadocJDKFiles)
     testCompile(javadocJDKFiles)
-    compile 'org.freemarker:freemarker:2.3.18'
+    compile 'org.freemarker:freemarker:2.3.23'
```
cmnbroad: @magicDGS Upgrading seems like a good idea. Since it sounds like there are some incompatibilities introduced, have you by any chance tried this with the gatkDoc process ? @jonn-smith Have you by any any chance tried using this version ? I doubt there is anything that would prevent us from upgrading, but it would be nice to know if there is any fallout.

jonn-smith: @cmnbroad I just tried a quick clean build - we'll need to make some code changes to make it compile (the deprecated warnings are causing the build to fail).

I think upgrading is a good idea, we'll just need to do some extra legwork to make sure we're using the most up-to-date interfaces in to FreeMarker (to fix the warnings) and I'm not sure what the full list is.

magicDGS: I install this branch in my local repository (it is not in the artifactory due to #80) and use it in the latest [gatk-master](https://github.com/broadinstitute/gatk/tree/2f2eb73be34e822ebea84b9f116c506768c625a1). Using `./gradlew clean gatkDoc` no error/warning was thrown.

Regarding nice features in this version that may be useful for barclay, and my main idea behind this change, are the built-in handling of special numbers ("INF", "-INF", "NaN") to check argument ranges (I could add this PR or a future PR to populate "NaN" instead of "NA" for some arguments).

cmnbroad: Ok, sounds good. I had tried running the same test last night so, this seems good to go.

magicDGS:
Thanks @cmnbroad!

PR #74, opened 2017-06-12
Fix json name

### LINKED ISSUE #72: JSON output file names should not include the output file extension
For example, if the output extension is "html", the JSON files are currently called "workunit.html.json". This behavior was carried over from GATK3, but is unnecessary and a little misleading. The work unit output format and extension are independent of JSON, so the JSON file should just be called "workunit.json".

magicDGS:
Closes #72


src/main/java/org/broadinstitute/barclay/help/DocWorkUnitHandler.java
```diff
@@ -56,6 +56,13 @@ public String getDestinationFilename(final DocWorkUnit workUnit) {
     }
 
     /**
+     * Returns the destination prefix (no paths) that the handler would like
```
cmnbroad: I think this comment is wrong - its the JSON output file name not the destination prefix.

src/main/java/org/broadinstitute/barclay/help/DocWorkUnitHandler.java
```diff
@@ -56,6 +56,13 @@ public String getDestinationFilename(final DocWorkUnit workUnit) {
...
+     */
+    public String getJsonFilename(final DocWorkUnit workUnit) {
```
cmnbroad: Can you change the names of these new methods (this one and the DocWorkUnit one) so they use "JSON " instead of "Json" since its an acronym.

src/main/java/org/broadinstitute/barclay/help/DocWorkUnitHandler.java
magicDGS: True, I don't know what I was thinking about... Changed!

src/main/java/org/broadinstitute/barclay/help/DocWorkUnitHandler.java
magicDGS: Done.

magicDGS:
I think that this is ready to go @cmnbroad once the tests pass! Thanks for reviewing!

cmnbroad:
Thanks!

magicDGS:
Thank you too for reviewing!

Issue #71, opened 2017-06-07
Test docgen code at a granular level


magicDGS:
From @cmnbroad (https://github.com/broadinstitute/barclay/pull/68#issuecomment-306683475):
> One high-level request is that we find a way to incrementally unit-test new functionality like this at a granular level. We currently only have coarse-grained, file-based integration tests, and I'd like to find a way to avoid proliferating a new set of test files with each new feature, in addition to changing many/all of the existing ones. @magicDGS any thoughts on how we can address that ?

magicDGS:
Regarding this issue, I can tell you what it is the idea that I'm trying to implement in my own code:

1. As in the current test implementation, I execute the javadoc to check that the whole doclet code is working and the templates are correct. In Barclay, I suggest to keep just the test for the default doclet/templates with some basic annotated classes, but remove the custom `TestDoclet` and associated templates.
2. For concrete features that may or may not be included in the basic doclet generation test, I'm trying to use a mocking framework (e.g. [Mockito](http://site.mockito.org/)) for `ClassDoc`. Like that, I can create `DocWorkUnit`s without the need of executing the javadoc, and testing methods for docgen classes in a granular way.

Is this reasonable for you, @cmnbroad?

cmnbroad:
Yes, a mocking framework is probably the right way to do this. Do you already have code that does this ? I'd like to solve this problem in Barclay before we add any more new features.

magicDGS:
I was exploring this a bit in my toolkit and I implemented some minor tests as a proof-of-concept. If you like, I can add a couple of tests with Mockito and remove the `TestDoclet` code by the end of next week. Let me know if that fits you!

cmnbroad:
@magicDGS That would be very useful. I wouldn't remove the TestDoclet test though, since we definitely need to test both default and custom doclet cases. At some point we may have enough mocked unit tests that we can remove one of them, but for now I think we should keep both. At least this will give us a start in the direction of less file bloat.

PR #75, opened 2017-06-14
Add tests using mocked objects

magicDGS:
As promised in #71


magicDGS:
Friendly ping here, @cmnbroad! I would like to have your approval for starting implementing more mocked tests and submit more PRs!

src/test/java/org/broadinstitute/barclay/help/DocGenBaseTest.java
```diff
@@ -0,0 +1,76 @@
...
+                clazz.getAnnotation(DocumentedFeature.class),
+                classDoc,
+                clazz
+        );
+    }
+
+    protected DocWorkUnit createDocWorkUnit(final Class<?> clazz) {
+        return createDocWorkUnit(clazz, emptyClassDoc());
+    }
+}
```
cmnbroad: I think it might be more useful for this class to something that is instantiated by unit test classes, rather than be a superclass for them. Then the tests could instantiate one of these (maybe passing in whatever doclet or workunit class they want), or just call static methods, and create appropriate mocks for the given test. This would make the tests totally self-contained (i.e., the testPropertyMap test method below would have everything it depends on defined locally, right in the method), and this class could then be used to write unit tests for, say, both the default doclet and the "test" doclet. At some point, we might even want to create another test doclet class to test some feature, and that doclet could also be passed to this class to use to create mocks. Just looking at this it doesn't seem that hard to do, and would be more flexible, but let me know if I'm missing something.

magicDGS:
I think that it is a good idea to separate testing and no-testing code, @cmnbroad. Following your suggestions with some modifications, I added three commits to this PR:

* Repackage test input classes: now the documented features/clps for testing live in `org.broadinstitute.barclay.help.testinputs`
* Repackage test doclet code: although you mention to keep the mocking test separate, I rather prefer to separate code not performing any test; that's why the `TestDoclet` and `TestDocWorkUnitHandler` now live in `org.broadinstitute.barclay.help.testdoclets`
* I refactor also the `DocGenBaseTest`to be an utility class for mocking (now called `DocGenMocking`.

---

*Note: I rebased the commits in the latest master to be sure that the changes does not affect the bash-completion code*

src/test/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandlerTest.java
```diff
@@ -0,0 +1,46 @@
...
+                {DocGenMocking.createDocWorkUnit(DEFAULT_DOC_WORK_UNIT_HANDLER, TestExtraDocs.class, DocGenMocking.mockClassDoc("Javadoc description \nin two lines", Collections.singletonMap("", "Inline tag"))),
+                        "Inline tag", "Javadoc description in two lines",
+                        "Test extra docs group name", ""}
+        };
+    }
+
+    @Test(dataProvider = "workUnitForDocStrings")
+    public void testGetDocumentationStrings(final DocWorkUnit workUnit,
+            final String description, final String summary,
+            final String groupName, final String groupSummary) throws Exception {
```
cmnbroad: all of these expected values should have names that are prefixed with "expected...".

src/test/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandlerTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+    private final static DefaultDocWorkUnitHandler DEFAULT_DOC_WORK_UNIT_HANDLER =
+            new DefaultDocWorkUnitHandler(new HelpDoclet());
+
+    @DataProvider
+    public Object[][] workUnitForDocStrings() throws Exception {
+        return new Object[][] {
+                {DocGenMocking.createDocWorkUnit(DEFAULT_DOC_WORK_UNIT_HANDLER, TestArgumentContainer.class),
+                        "", "Argument container class for testing documentation generation.",
+                        "Test feature group name", "Test program group used for testing"},
```
cmnbroad: It took me a little hunting to figure where these strings came from. They should be factored out into symbolic constants (probably defined in the test input class) that can be used both here and in the test input class. Then its easy to see why they're the expected values, and they'll be prefixed with, say "TestArgumentContainer." so they'll be easily navigable in the IDE.

src/test/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandlerTest.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    @Test(dataProvider = "workUnitForDocStrings")
+    public void testGetDocumentationStrings(final DocWorkUnit workUnit,
+            final String description, final String summary,
+            final String groupName, final String groupSummary) throws Exception {
+        Assert.assertEquals(DEFAULT_DOC_WORK_UNIT_HANDLER.getDescription(workUnit), description);
+        Assert.assertEquals(DEFAULT_DOC_WORK_UNIT_HANDLER.getSummaryForWorkUnit(workUnit), summary);
+        Assert.assertEquals(DEFAULT_DOC_WORK_UNIT_HANDLER.getGroupNameForWorkUnit(workUnit), groupName);
+        Assert.assertEquals(DEFAULT_DOC_WORK_UNIT_HANDLER.getGroupSummaryForWorkUnit(workUnit), groupSummary);
```
cmnbroad: I think we ultimately want to shoot for a structure with (at least) one unit test per method, and symmetry between the methods and test methods, even if we don't have complete unit test coverage now. This test method should be replaced with 4 separate test methods. That would also cut down on the number of "expected" arguments for each test method.

You may need to produce a separate data provider for each test method (like this test, which uses two work units, one from TestArgumentContainer and one from TestExtraDocs, each with the corresponding expected value for the data entry, but thats ok.

cmnbroad: Also, these unit tests are really helpful to point out flaws in the underlying code - I never fully refactored the code I ported form GATK3 and it shows. All of the methods being tested here are DocWorkUnitHandler methods, but every single one of them takes a workUnit argument, and the implementations only use the work unit itself. These should all be methods on DocWorkUnit! But thats for another day. Still, these kinds of tests are really helpful since they make that easy to see.

cmnbroad:
@magicDGS Also, on this:

> I rather prefer to separate code not performing any test; that's why the TestDoclet and TestDocWorkUnitHandler now live in org.broadinstitute.barclay.help.testdoclets

I think its fine for now, but once we start to accumulate more mocks and tests, I may want to separate them as I suggested. But we can hold off on that.

src/test/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandlerTest.java
```diff
@@ -0,0 +1,46 @@
...
+import org.testng.Assert;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import java.util.Collections;
+
+/**
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public class DefaultDocWorkUnitHandlerTest {
```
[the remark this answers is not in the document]
magicDGS: Done.

src/test/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandlerTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+    private final static DefaultDocWorkUnitHandler DEFAULT_DOC_WORK_UNIT_HANDLER =
+            new DefaultDocWorkUnitHandler(new HelpDoclet());
+
+    @DataProvider
+    public Object[][] workUnitForDocStrings() throws Exception {
```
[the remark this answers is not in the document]
magicDGS: I use to add them in every test, in case that methods are changed to throw an exception and it is missing. Removed here.

src/test/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandlerTest.java
magicDGS: Done.

src/test/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandlerTest.java
```diff
@@ -0,0 +1,46 @@
...
+        return new Object[][] {
+                {DocGenMocking.createDocWorkUnit(DEFAULT_DOC_WORK_UNIT_HANDLER, TestArgumentContainer.class),
+                        "", "Argument container class for testing documentation generation.",
+                        "Test feature group name", "Test program group used for testing"},
+                {DocGenMocking.createDocWorkUnit(DEFAULT_DOC_WORK_UNIT_HANDLER, TestExtraDocs.class, DocGenMocking.mockClassDoc("Javadoc description")),
+                        "", "Javadoc description",
+                        "Test extra docs group name", ""},
+                // TODO: because DefaultWorkUnitHandler does not have an extra tag, the mocked class doc uses an empty tag name for the inline tag
```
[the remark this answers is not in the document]
magicDGS: Done. And sorry for my clunky English!

src/test/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandlerTest.java
```diff
@@ -0,0 +1,46 @@
...
+                // TODO: perhaps Barclay should include a default inline tag prefix for being able to re-use between different toolkits
```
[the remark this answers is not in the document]
magicDGS: I just thought that it will be good to have tag support out-of-the-box, using the default helpdoclet. But I do not need it...

Removed TODO

src/test/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandlerTest.java
magicDGS: Done.

src/test/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandlerTest.java
magicDGS: Refactor to have one test per method.

src/test/java/org/broadinstitute/barclay/help/DocGenMocking.java
```diff
@@ -0,0 +1,73 @@
...
+import org.mockito.Mockito;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Map;
+
+/**
+ * @author Daniel Gomez-Sanchez (magicDGS)
+ */
+public class DocGenMocking {
```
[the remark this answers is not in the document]
magicDGS: Done.

src/test/java/org/broadinstitute/barclay/help/DocGenMocking.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    /**
+     * Generates an ClassDoc with the javadoc text
+     */
+    public static ClassDoc mockClassDoc(final String javadocText) {
+        return mockClassDoc(javadocText, Collections.emptyMap());
+    }
+
+
```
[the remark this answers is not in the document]
magicDGS: Removed.

src/test/java/org/broadinstitute/barclay/help/DocGenMocking.java
```diff
@@ -0,0 +1,73 @@
...
+                .toArray(Tag[]::new);
+        Mockito.when(mockedClassDoc.firstSentenceTags()).thenReturn(javadoc);
+        final Tag[] inline = inlineTags.entrySet().stream()
+                .map(entry -> mockTag(entry.getKey(), entry.getValue()))
+                .toArray(Tag[]::new);
+        Mockito.when(mockedClassDoc.inlineTags()).thenReturn(inline);
+        return mockedClassDoc;
+    }
+
+
```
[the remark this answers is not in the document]
magicDGS: Removed.

magicDGS:
I addressed your comments, @cmnbroad -back to you and thanks for reviewing.

src/test/java/org/broadinstitute/barclay/help/DocGenMocks.java
```diff
@@ -0,0 +1,71 @@
...
+                docWorkUnitHandler,
+                clazz.getAnnotation(DocumentedFeature.class),
+                classDoc,
+                clazz
+        );
+    }
+
+    public static DocWorkUnit createDocWorkUnit(final DocWorkUnitHandler docWorkUnitHandler, final Class<?> clazz) {
+        return createDocWorkUnit(docWorkUnitHandler, clazz, emptyClassDoc());
+    }
```
cmnbroad: These last two methods don't seem to be mock related; my preference to see the DocWorkUnit creation code inlined at the calls sites to make it clear whats going on.

build.gradle
```diff
@@ -59,6 +59,7 @@ dependencies {
     compile 'com.google.code.gson:gson:2.2.2'
 
     testCompile 'org.testng:testng:6.9.6'
+    testCompile 'org.mockito:mockito-core:2.8.9'
```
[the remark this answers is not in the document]
magicDGS: Done.

src/test/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandlerUnitTest.java
```diff
@@ -0,0 +1,88 @@
...
+                {DocGenMocks.createDocWorkUnit(getNewInstance(), TestArgumentContainer.class),
+                        ""},
+                {DocGenMocks.createDocWorkUnit(getNewInstance(), TestExtraDocs.class, DocGenMocks.mockClassDoc("Javadoc description")),
+                        ""},
+                // because DefaultWorkUnitHandler does not define an extra tag, the mocked class doc uses an empty tag name for the inline tag
+                {DocGenMocks.createDocWorkUnit(getNewInstance(), TestExtraDocs.class, DocGenMocks
+                        .mockClassDoc("Javadoc description \nin two lines", Collections.singletonMap("", "Inline tag"))),
+                        "Inline tag"}
+        };
+    }
```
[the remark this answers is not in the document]
magicDGS: Done.

src/test/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandlerUnitTest.java
```diff
@@ -0,0 +1,88 @@
...
+    public void testGetGroupNameForWorkUnit(final DocWorkUnit workUnit, final String expectedDescription) {
+        Assert.assertEquals(getNewInstance().getGroupNameForWorkUnit(workUnit), expectedDescription);
+    }
+
+    @Test(dataProvider = "workUnitGroupSummary")
+    public void testGetGroupSummaryForWorkUnit(final DocWorkUnit workUnit, final String expectedDescription) {
+        Assert.assertEquals(getNewInstance().getGroupSummaryForWorkUnit(workUnit), expectedDescription);
+    }
+
+    private static DefaultDocWorkUnitHandler getNewInstance() {
```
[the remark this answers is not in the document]
magicDGS: Done.

src/test/java/org/broadinstitute/barclay/help/DocGenMocks.java
magicDGS: So are you suggesting to remove completely this methods? I think that they are useful, because that code is gonna be repeated all around the mocked tests...Maybe an utility class for testing the doclet?

cmnbroad: Both of these (as well as emptyClassDoc and mockClassDoc(String)) are just wrappers around a single call, and to me only obscure whats being mocked and what isn't. I'd strongly prefer to see them removed and inlined now. Maybe as we develop more real tests we can factor stuff out.

src/test/java/org/broadinstitute/barclay/help/DocWorkUnitTest.java
```diff
@@ -0,0 +1,61 @@
...
+    @Test
+    public void testPropertyMap() {
+        final DocWorkUnit workUnit = DocGenMocks
+                .createDocWorkUnit(getDefaultDocWorkUnitHandler(), TestExtraDocs.class);
+        Assert.assertTrue(workUnit.getRootMap().isEmpty());
+        Assert.assertEquals(workUnit.getProperty("test_property"), null);
+        workUnit.setProperty("test_property", "test_value");
+        Assert.assertEquals(workUnit.getProperty("test_property"), "test_value");
+        Assert.assertEquals(workUnit.getRootMap().size(), 1);
+    }
```
cmnbroad: The "test_property" and "test_value" strings should be factored out into constants.

src/test/java/org/broadinstitute/barclay/help/DocWorkUnitTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Test
+    public void tesGetBetaFeature() {
+        Assert.assertFalse(
+                DocGenMocks.createDocWorkUnit(getDefaultDocWorkUnitHandler(), TestExtraDocs.class).getBetaFeature());
+        Assert.assertTrue(
+                DocGenMocks.createDocWorkUnit(getDefaultDocWorkUnitHandler(), TestArgumentContainer.class).getBetaFeature());
+    }
+
```
cmnbroad: Its preferable to use a data provider for these methods wherever possible, instead of repeating the asserts for each test case.

src/test/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandlerUnitTest.java
```diff
@@ -0,0 +1,89 @@
...
+                {TestArgumentContainer.class, TestArgumentContainer.GROUP_NAME},
+                {TestExtraDocs.class, TestExtraDocs.GROUP_NAME}
+        };
+    }
+
+    @DataProvider
+    public static Object[][] workUnitGroupSummary() {
+        return new Object[][] {
+                {TestArgumentContainer.class, TestProgramGroup.DESCRIPTION},
+                {TestExtraDocs.class, Collections.EMPTY_MAP, ""},
```
cmnbroad: I know you're still in the middle of this, but one has two params and one has three.

src/test/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandlerUnitTest.java
```diff
@@ -0,0 +1,89 @@
...
+            final String expectedSummary) {
+        final DocWorkUnit mockWorkUnit = createMockWorkUnit(docWorkUnitClazz, javadocText, inlineTags);
+        Assert.assertEquals(getDefaultWorkUnitHandlerInstance().getSummaryForWorkUnit(mockWorkUnit), expectedSummary);
+    }
+
+    @Test(dataProvider = "workUnitGroupName")
+    public void testGetGroupNameForWorkUnit(final Class<?> docWorkUnitClazz, final String expectedGroupName) {
+        final DocWorkUnit mockWorkUnit = DocGenMocks.createDocWorkUnit(
+                getDefaultWorkUnitHandlerInstance(), docWorkUnitClazz);
+        Assert.assertEquals(getDefaultWorkUnitHandlerInstance().getGroupNameForWorkUnit(mockWorkUnit), expectedGroupName);
```
cmnbroad: There are several of these where there are multiple calls to getDefaultWorkUnitHandlerInstance in the same test. It think there should be one DefaultWorkUnitHandler created here, which is used to both mock the workUnit and to retrieve the groupName.

The fact that this test works as is indicates that getGroupNameForWorkUnit (and getGroupSummaryForWorkUnit ) should really be methods on DocWorkUnit, not on the handler, since it appears to not matter what handler instance is used. (I think this can up in a previous PR too). But thats for another day. For now we should fix the tests to use the same one.

src/test/java/org/broadinstitute/barclay/help/DocGenMocks.java
magicDGS: Done, although I really think that the shortcuts are useful for the future, if we keep adding more complexity to the mocked objects.

src/test/java/org/broadinstitute/barclay/help/DocWorkUnitTest.java
magicDGS: Done.

src/test/java/org/broadinstitute/barclay/help/DocWorkUnitTest.java
magicDGS: There are only two possibilities in this case, and in principle they shouldn't change. Anyway, I refactored...

src/test/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandlerUnitTest.java
magicDGS: My bad, a refactoring error. Maybe we should add a test for data providers as in HTSJDK/Picard/GATK (in a different PR).

src/test/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandlerUnitTest.java
magicDGS: I think that it shouldn't be re-used, because we should test how a state-less WorkUnitHandler behaves.

Regarding your side note, the default handler is independent of the instance used, but the user can override the behaviour. An usage case is to document a toolkit which includes another one (e.g., Picard inside GATK). In that case, a custom handler may want to group together all the Picard programs to have them in a different page or pointing that they do not belong to the custom toolkit but to a third-party. I think that this methods belongs to the handle even if the default behaviour is instance-independent...

magicDGS: No action was taken here.

cmnbroad: @magicDGS The point I was making was that using a different handler to get attributes like group name than was originally used to create the work unit isn't how Barclay itself gets them, so write the test this way ? There isn't anything that says that a handler needs to be stateless - the fact that this currently works with the default handler is completely accidental. It would be quite easy to make a perfectly valid change to the default handler that would cause this test to fail, but pass if the same handler was used.

The doc should probably be more explicit about this, but in the meantime it seems much more useful, and less brittle, to have a test that ensures that this method does the right thing when used as Barclay uses it.

magicDGS: Done.

magicDGS:
Back to you @cmnbroad!

cmnbroad:
@magicDGS Thanks for doing this!

magicDGS:
Thanks for the review, @cmnbroad!

PR #78, opened 2017-06-19
Extract method to allow custom doclet options

magicDGS:
Very simple patch to allow integrating custom options for a doclet. Currently parsing custom options is a bit painful, but after this change it is as simple as: 

```java
public class CustomDoclet extends HelpDoclet {
    
    ....

    private static final String DEPENDENCY_VERSION_OPTION = "-dependency-version";
    protected static Map<String, String> dependencyVersions = new LinkedHashMap<>();

    @Override
    protected parseOption(Option[] options) {
        super.parseOption(options);
        if (options[0].equals(DEPENDENCY_VERSION_OPTION))
            dependencyVersions.put(options[1], options[2]);
    }

    public static int optionLength(final String option) {
         if (option.equals(DEPENDENCY_VERSION_OPTION)) return 3;
         return HelpDoclet.optionLength(option);
    }

    ....
}
```

### DIFF
--- a/src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
+++ b/src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
@@ -117,13 +117,16 @@ public class HelpDoclet {
     }
 
     /**
-     * Parse the options for the javadoc command line. The first item is the option name, and the
-     * following the number of items returned by {@link #optionLength(String)} for that option.
+     * Handles javadoc command line options. The first entry in the {@code options} array is the
+     * option name. Subsequent entries contain the option's arguments, the number of which is
+     * determined by the value returned from {@link #optionLength(String)} for that option.
      *
-     * Implementations should override this method and call it to parse the required options.
-     * In addition, to support custom options {@link #optionLength(String)} may be implemented.
+     * <p>Custom Barclay doclets that want to have custom command line options should override this
+     * method, and also provide an implementation of the static method {@link #optionLength}).
+     * Both methods should handle the custom options, and then delegate back to the base class
+     * methods defined here to allow default handling of builtin options.
      *
-     * @param options
+     * @param options Options to parse.
      */
     protected void parseOption(final String[] options) {
         if (options[0].equals(SETTINGS_DIR_OPTION))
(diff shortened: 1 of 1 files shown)


magicDGS:
A very simple patch, @cmnbroad - mostly copy-paste code to extract a method and javadoc updates to explain how to implement the custom options...

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
```diff
@@ -132,8 +117,40 @@ else if (!settingsDir.isDirectory())
     }
 
     /**
+     * Parse the options for the javadoc command line. The first item is the option name, and the
+     * following the number of items returned by {@link #optionLength(String)} for that option.
```
cmnbroad: Change this comment to something like:

Handle javadoc command line options. The first entry in the {@code options} array is the option name. Subsequent entries contain the option's arguments, the number of which is determined by the value returned from {@link #optionLength(String)} for that option.

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
```diff
@@ -132,8 +117,40 @@ else if (!settingsDir.isDirectory())
...
+     *
+     * Implementations should override this method and call it to parse the required options.
+     * In addition, to support custom options {@link #optionLength(String)} may be implemented.
```
cmnbroad: Change this to:

Custom Barclay doclets that want to have custom command line options should override this method, and also provide an implementation of the static method {@link #optionLength}). Both methods should handle the custom options, and then delegate back to the base class methods defined here to allow default handling of builtin options.

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
```diff
@@ -132,8 +117,40 @@ else if (!settingsDir.isDirectory())
...
+        }
+    }
+
+    /**
      * Validate the given options against options supported by this doclet.
      *
+     * <p>Note: for custom options, doclets should implement a similar static method and call the
+     * one from {@link HelpDoclet} to maintain required options. In addition, custom options should
+     * be handled by {@link #parseOption(String[])}.
+     *
```
cmnbroad: Change this comment to:

Custom Barclay doclets that want to have custom command line options should provide an implementation of this static method, and also override override {@link parseOption}). Both methods should handle the custom options, delegating back to the base class methods defined here to allow default handling of builtin options.

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java

magicDGS:
Addressed comments. Back to yo, @cmnbroad

magicDGS:
Thanks @cmnbroad!

PR #82, opened 2017-06-23
Updates to barclay to support tab-completion and other output files

jonn-smith:
Here are some updates to barclay that will support generation of tab-completion files for bash (tested on OSX 10.12.5 - Sierra on GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin16))

It includes new templates and expected values for generated files based on test classes.  It also includes one or two minor bug fixes.

I needed to add more information to the parameters map that gets passed to FreeMarker, so I added a few more fields to it.

Some private methods were downgraded to protected to expose some functionality that was necessary.

### DIFF
--- a/src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
+++ b/src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
@@ -1,0 +1,548 @@
+package org.broadinstitute.barclay.help;
+
+import com.sun.javadoc.RootDoc;
+import freemarker.template.Configuration;
+import freemarker.template.Template;
+import freemarker.template.TemplateException;
+
+import java.io.*;
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * For testing of Bash tab completion generation.
+ *
+ * Using this to generate tab-completion files requires that there is a
+ * wrapper script around the call to java that acts as a user-interface.
+ *
+ * This is required because of how Bash handles tab completion - it keys off
+ * of the first word typed in a line.  When invoking directly from java, Bash
+ * will complete for the `java` command, but will not know how to complete for a
+ * jar incorporating Barclay-enabled arguments.
+ *
+ * This is a known issue and is being investigated for remedies in the future.
+ */
+public class BashTabCompletionDoclet extends HelpDoclet {
+
+    // Barclay BashTabCompletionDoclet custom Command-line Arguments:
+
+    // All these arguments are optional, but it is highly recommended
+    // to specify the caller script name.
+
+    final private static String CALLER_SCRIPT_NAME = "-caller-script-name";
+
+    final private static String CALLER_SCRIPT_PREFIX_LEGAL_ARGS          = "-caller-pre-legal-args";
+    final private static String CALLER_SCRIPT_PREFIX_ARG_VALUE_TYPES     = "-caller-pre-arg-val-types";
+    final private static String CALLER_SCRIPT_PREFIX_MUTEX_ARGS          = "-caller-pre-mutex-args";
+    final private static String CALLER_SCRIPT_PREFIX_ALIAS_ARGS          = "-caller-pre-alias-args";
+    final private static String CALLER_SCRIPT_PREFIX_ARG_MIN_OCCURRENCES = "-caller-pre-arg-min-occurs";
+    final private static String CALLER_SCRIPT_PREFIX_ARG_MAX_OCCURRENCES = "-caller-pre-arg-max-occurs";
+
+    final private static String CALLER_SCRIPT_POSTFIX_LEGAL_ARGS          = "-caller-post-legal-args";
+    final private static String CALLER_SCRIPT_POSTFIX_ARG_VALUE_TYPES     = "-caller-post-arg-val-types";
+    final private static String CALLER_SCRIPT_POSTFIX_MUTEX_ARGS          = "-caller-post-mutex-args";
+    final private static String CALLER_SCRIPT_POSTFIX_ALIAS_ARGS          = "-caller-post-alias-args";
+    final private static String CALLER_SCRIPT_POSTFIX_ARG_MIN_OCCURRENCES = "-caller-post-arg-min-occurs";
+    final private static String CALLER_SCRIPT_POSTFIX_ARG_MAX_OCCURRENCES = "-caller-post-arg-max-occurs";
+
+    // =============================================
+
+    // Variables that are set on the command line when running this doclet:
+
+    /**
+     * Name of the executable / wrapper script that will actually invoke the java process.
+     * This wrapper script would call into the JAR and tell it which class to run.
+     */
+    private String callerScriptName = null;
+
+    /**
+     * Arguments to the executable / wrapper script that come before any Java class names / tools.
+     *
+     * This is expected to be a space-delimited string with the options themselves as they should be
+     * typed by the user.
+     *
+     * This syntax is used to pass this information to directly to the bash completion script.
+     *
+     * Example: {@code "--help --info --list --inputFile --outFolder --memSize --multiplier"}
+     */
+    private String callerScriptPrefixLegalArgs       = "";
+
+    /**
+     * Types of the arguments that the executable / wrapper script is expecting before any Java class names / tools.
+     * The order of these space-delimited types should correspond to the contents of {@link #callerScriptPrefixLegalArgs}
+     *
+     * This is expected to be a space-delimited string of types.
+     *
+     * Currently accepted type values are the following (not case-sensitive):
+     *
+     *      {@code file}
+     *      {@code folder}
+     *      {@code directory}
+     *      {@code int}
+     *      {@code long}
+     *      {@code double}
+     *      {@code float}
+     *      {@code null}   (to be used in the case of an argument that acts as a flag [i.e. one that takes no additional input])
+     *
+     * This syntax is used to pass this information to directly to the bash completion script.
+     *
+     * Example: {@code "null null null file directory int double"}
+     */
+    private String callerScriptPrefixArgValueTypes   = "";
+
+    /**
+     * Sets of arguments to the executable / wrapper script that are mutually exclusive to each other and
+     * are expected before any Java class names / tools.
+     *
+     * This is expected to be a string with mutex information for each argument that is mutually exclusive with another
+     * argument.  The format for this string is:
+     *
+     * {@code FOO;mutexToFoo1[,mutexToFoo2][,mutexToFoo3]... BAR;mutexToBar1[,mutexToBar2][,mutexToBar3]... }
+     *
+     *  where:
+     *
+     * {@code FOO} is an argument to the wrapper script which is expected before any Java class names / tools
+     * {@code mutexToFoo1} is an argument with which {@code FOO} is mutually exclusive without leading decorators (usually - or --)
+     * {@code mutexToFoo2} is an argument with which {@code FOO} is mutually exclusive without leading decorators (usually - or --)
+     * {@code mutexToFoo3} is an argument with which {@code FOO} is mutually exclusive without leading decorators (usually - or --)
+     *  and
+     * {@code BAR} is an argument to the wrapper script which is expected before any Java class names / tools
+     * {@code mutexToBar1} is an argument with which {@code BAR} is mutually exclusive without leading decorators (usually - or --)
+     * {@code mutexToBar2} is an argument with which {@code BAR} is mutually exclusive without leading decorators (usually - or --)
+     * {@code mutexToBar3} is an argument with which {@code BAR} is mutually exclusive without leading decorators (usually - or --)
+     *
+     * This can be thought of as a set of such argument relationships and does not have any ordering scheme.
+     *
+     * This syntax is used to pass this information to directly to the bash completion script.
+     *
+     * Example: {@code "--help;info,list,inputFile --info;help,list,inputFile"}
+     */
+    private String callerScriptPrefixMutexArgs       = "";
+
+    /**
+     * Sets of arguments to the executable / wrapper script that are aliases of each other and
+     * are expected before any Java class names / tools.
+     * For example, full argument names and short names for those arguments.
+     *
+     * This is expected to be a string with alias information for each argument that is an alias of another
+     * argument.  The format for this string is:
+     *
+     * {@code FOO;aliasToFoo1[,aliasToFoo2][,aliasToFoo3]... BAR;aliasToBar1[,aliasToBar2][,aliasToBar3]... }
+     *
+     *  where:
+     *
+     * {@code FOO} is an argument to the wrapper script which is expected before any Java class names / tools
+     * {@code aliasToFoo1} is an argument which is an alias to {@code FOO}
+     * {@code aliasToFoo2} is an argument which is an alias to {@code FOO}
+     * {@code aliasToFoo3} is an argument which is an alias to {@code FOO}
+     *  and
+     * {@code BAR} is an argument to the wrapper script which is expected before any Java class names / tools
+     * {@code aliasToBar1} is an argument which is an alias to {@code BAR}
+     * {@code aliasToBar2} is an argument which is an alias to {@code BAR}
+     * {@code aliasToBar3} is an argument which is an alias to {@code BAR}
+     *
+     * This can be thought of as a set of such argument relationships and does not have any ordering scheme.
+     *
+     * This syntax is used to pass this information to directly to the bash completion script.
+     *
+     * Example: {@code "--help;-h --info;-i --inputFile;-if,-infile,-inny"}
+     */
+    private String callerScriptPrefixAliasArgs       = "";
+
+    /**
+     * The minimum number of occurrences of each argument that the executable / wrapper script is expecting
+     * before any Java class names / tools.
+     * This is expected to be a space-delimited string with the min occurrences as {@code integer} values.
+     *
+     * The order of these space-delimited values should correspond to the contents of {@link #callerScriptPrefixLegalArgs}
+     *
+     * This is used in the logic that tracks the number of times an option is specified.
+     *
+     * This syntax is used to pass this information to directly to the bash completion script.
+     *
+     * Example: {@code "0 0 0 1 1 0 0 0"}
+     */
+    private String callerScriptPrefixMinOccurrences  = "";
+
+    /**
+     * The maximum number of occurrences of each argument that the executable / wrapper script is expecting
+     * before any Java class names / tools.
+     * This is expected to be a space-delimited string with the max occurrences as {@code integer} values.
+     *
+     * The order of these space-delimited values should correspond to the contents of {@link #callerScriptPrefixLegalArgs}
+     *
+     * This is used in the logic that tracks the number of times an option is specified.
+     *
+     * This syntax is used to pass this information to directly to the bash completion script.
+     *
+     * Example: {@code "1 1 1 1 1 1 1 1"}
+     */
+    private String callerScriptPrefixMaxOccurrences  = "";
+
+
+    /**
+     * Arguments to the executable / wrapper script that come after any Java class names / tools.  The start of these
+     * options is indicated by the user inputting the special option {@code --}
+     *
+     * The format of this variable is identical to {@link #callerScriptPrefixLegalArgs}
+     */
+    private String callerScriptPostfixLegalArgs      = "";
+
+    /**
+     * Types of the arguments that the executable / wrapper script is expecting after any Java class
+     * names / tools.  The start of these options is indicated by the user inputting the special option {@code --}
+     *
+     * The order of these space-delimited types should correspond to the contents of {@link #callerScriptPostfixLegalArgs}
+     *
+     * The format of this variable is identical to {@link #callerScriptPrefixArgValueTypes}
+     */
+    private String callerScriptPostfixArgValueTypes  = "";
+
+    /**
+     * Sets of arguments to the executable / wrapper script that are mutually exclusive to each other and
+     * are expected after any Java class names / tools.  The start of these options is indicated by the user
+     * inputting the special option {@code --}
+     *
+     * The format of this variable is identical to {@link #callerScriptPrefixMutexArgs}
+     */
+    private String callerScriptPostfixMutexArgs      = "";
+
+    /**
+     * Sets of arguments to the executable / wrapper script that are aliases of each other and
+     * are expected after any Java class names / tools.  The start of these options is indicated by the user
+     * inputting the special option {@code --}
+     *
+     * The format of this variable is identical to {@link #callerScriptPrefixAliasArgs}
+     */
+    private String callerScriptPostfixAliasArgs      = "";
+
+    /**
+     * The minimum number of occurrences of each argument that the executable / wrapper script is expecting
+     * after any Java class names / tools.  The start of these options is indicated by the user
+     * inputting the special option {@code --}
+     *
+     * The order of these space-delimited types should correspond to the contents of {@link #callerScriptPostfixLegalArgs}
+     *
+     * The format of this variable is identical to {@link #callerScriptPrefixMinOccurrences}
+     */
+    private String callerScriptPostfixMinOccurrences = "";
+
+    /**
+     * The maximum number of occurrences of each argument that the executable / wrapper script is expecting
+     * after any Java class names / tools.  The start of these options is indicated by the user
+     * inputting the special option {@code --}
+     *
+     * The order of these space-delimited types should correspond to the contents of {@link #callerScriptPostfixLegalArgs}
+     *
+     * The format of this variable is identical to {@link #callerScriptPrefixMaxOccurrences}
+     */
+    private String callerScriptPostfixMaxOccurrences = "";
+
+    /**
+     * True if the executable / wrapper script has arguments that are expected after any Java class names / tools.
+     * The start of these options is indicated by the user inputting the special option {@code --}
+     * The value of this is set internally based on the contents of {@link #callerScriptPostfixLegalArgs}
+     */
+    private boolean hasCallerScriptPostfixArgs       = false;
+
+    // =============================================
+
+    // Member variables:
+    protected static String outputFileExtension = "sh";
+    protected static String indexFileExtension = "sh";
+
+    // =============================================
+
+    public static boolean start(RootDoc rootDoc) {
+        try {
+            return new BashTabCompletionDoclet().startProcessDocs(rootDoc);
+        } catch (IOException e) {
+            throw new DocException("Exception processing javadoc", e);
+        }
+    }
+
+    private String quoteEachWord(final String sentence) {
+        return quoteEachWord(sentence, " ");
+    }
+    private String quoteEachWord(final String sentence, final String sep) {
+
+        return Stream.of(sentence.split(sep))
+                .map(s -> String.format("\"%s\"", s))
+                .collect(Collectors.joining(sep));
+    }
+
+    // Must add options that are applicable to this doclet so that they will be accepted.
+    public static int optionLength(final String option) {
+
+        if (option.equals(CALLER_SCRIPT_NAME) ||
+            option.equals(CALLER_SCRIPT_PREFIX_LEGAL_ARGS) ||
+            option.equals(CALLER_SCRIPT_PREFIX_ARG_VALUE_TYPES) ||
+            option.equals(CALLER_SCRIPT_PREFIX_MUTEX_ARGS) ||
+            option.equals(CALLER_SCRIPT_PREFIX_ALIAS_ARGS) ||
+            option.equals(CALLER_SCRIPT_PREFIX_ARG_MIN_OCCURRENCES) ||
+            option.equals(CALLER_SCRIPT_PREFIX_ARG_MAX_OCCURRENCES) ||
+            option.equals(CALLER_SCRIPT_POSTFIX_LEGAL_ARGS) ||
+            option.equals(CALLER_SCRIPT_POSTFIX_ARG_VALUE_TYPES) ||
+            option.equals(CALLER_SCRIPT_POSTFIX_MUTEX_ARGS) ||
+            option.equals(CALLER_SCRIPT_POSTFIX_ALIAS_ARGS) ||
+            option.equals(CALLER_SCRIPT_POSTFIX_ARG_MIN_OCCURRENCES) ||
+            option.equals(CALLER_SCRIPT_POSTFIX_ARG_MAX_OCCURRENCES) )
+        {
+            return 2;
+        }
+        else {
+            return HelpDoclet.optionLength(option);
+        }
+    }
+
+    @Override
+    protected void validateDocletStartingState() {
+        if ( callerScriptName == null ) {
+            // The user did not specify the caller script name.
+            // We cannot function under these conditions:
+            throw new RuntimeException("ERROR: You must specify a caller script name using the option: " + CALLER_SCRIPT_NAME);
+        }
+    }
+
+    @Override
+    protected boolean parseOption(final String[] option) {
+
+        // Do the stuff that HelpDoclet needs:
+        boolean hasParsedOption = super.parseOption(option);
+
+        if ( !hasParsedOption ) {
+
+            // Do the stuff that BashTabCompletionDoclet needs:
+            // (Yes this is inefficient because the parent class cycles through input args too.)
+            if (option[0].equals(CALLER_SCRIPT_NAME)) {
+
+                // Remove the last period and anything after it:
+                final int lastDotIndex = option[1].lastIndexOf('.');
+                if ( lastDotIndex != -1 ) {
+                    callerScriptName = option[1].substring(0, lastDotIndex);
+                }
+                else {
+                    callerScriptName = option[1];
+                }
+
+                hasParsedOption = true;
+            }
+
+            else if (option[0].equals(CALLER_SCRIPT_PREFIX_LEGAL_ARGS)) {
+                callerScriptPrefixLegalArgs = option[1];
+                hasParsedOption = true;
+            }
+            else if (option[0].equals(CALLER_SCRIPT_PREFIX_ARG_VALUE_TYPES)) {
+                // We have to format this option to contain quotes around each word:
+                callerScriptPrefixArgValueTypes = quoteEachWord(option[1]);
+                hasParsedOption = true;
+            }
+            else if (option[0].equals(CALLER_SCRIPT_PREFIX_MUTEX_ARGS)) {
+                // We have to format this option to contain quotes around each group of options:
+                callerScriptPrefixMutexArgs = quoteEachWord(option[1]);
+                hasParsedOption = true;
+            }
+            else if (option[0].equals(CALLER_SCRIPT_PREFIX_ALIAS_ARGS)) {
+                // We have to format this option to contain quotes around each group of options:
+                callerScriptPrefixAliasArgs = quoteEachWord(option[1]);
+                hasParsedOption = true;
+            }
+            else if (option[0].equals(CALLER_SCRIPT_PREFIX_ARG_MIN_OCCURRENCES)) {
+                callerScriptPrefixMinOccurrences = option[1];
+                hasParsedOption = true;
+            }
+            else if (option[0].equals(CALLER_SCRIPT_PREFIX_ARG_MAX_OCCURRENCES)) {
+                callerScriptPrefixMaxOccurrences = option[1];
+                hasParsedOption = true;
+            }
+
+            else if (option[0].equals(CALLER_SCRIPT_POSTFIX_LEGAL_ARGS)) {
+                callerScriptPostfixLegalArgs = option[1];
+                hasCallerScriptPostfixArgs = !callerScriptPostfixLegalArgs.isEmpty();
+                hasParsedOption = true;
+            }
+            else if (option[0].equals(CALLER_SCRIPT_POSTFIX_ARG_VALUE_TYPES)) {
+                // We have to format this option to contain quotes around each word:
+                callerScriptPostfixArgValueTypes = quoteEachWord(option[1]);
+                hasParsedOption = true;
+            }
+            else if (option[0].equals(CALLER_SCRIPT_POSTFIX_MUTEX_ARGS)) {
+                // We have to format this option to contain quotes around each word:
+                callerScriptPostfixMutexArgs = quoteEachWord(option[1]);
+                hasParsedOption = true;
+            }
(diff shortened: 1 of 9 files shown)


vdauwera:
This is awesome! What's the range of things that will autocomplete? Tool names and arguments? Any chance it also understands enum values? (as we say in French, give me your hand and I will grab your arm...)

jonn-smith:
Right now it's just tool names and arguments (as well as arguments to the launcher script itself).  I thought about enums and didn't put them in for now. It should be possible, but it'll take some additional Barclay code changes.

Also, it makes a few assumptions about argument value types (for instance, if it gets confused, it defaults to file tab-completion).

vdauwera:
Ok, good to know -- we'll want to document that somewhere. In fact it would be swell if you could add some explanation of this feature (as regards what a user would need to know) to the README. 

Thanks for doing this!

src/test/resources/org/broadinstitute/barclay/help/templates/BashTestDoclet/bash-completion.template.ftl
```diff
@@ -0,0 +1,343 @@
+
+####################
+# Tab completion file to allow for easy use of this tool with the command-line (bash/zsh)
+####################
+
+<#include "bash-completion.macros.ftl"/>
+
+####################################################################################################
+
+# High-level caller/dispatch script information:
```
magicDGS: I prefer to have in Barclay something that is not gatk specific, for being able to use the default template in other toolkits as a start point without exploring the templates...

src/main/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandler.java
```diff
@@ -667,8 +674,8 @@ private FieldDoc getFieldDoc(final ClassDoc classDoc, final String name, final b
      *         synonymous name.
      */
     private Pair<String, String> displayNames(String s1, String s2) {
-        s1 = s1 == null ? null : "-" + s1;
-        s2 = s2 == null ? null : "--" + s2;
+        s1 = ((s1 == null) || (s1.length() == 0)) ? null : "-" + s1;
```
magicDGS: Thanks for this, this is a bug that I found the other day in my json!

src/test/java/org/broadinstitute/barclay/help/DocumentationGenerationIntegrationTest.java
```diff
@@ -68,28 +68,45 @@
                 {HelpDoclet.class,
                         new File(inputResourcesDir + "helpTemplates/"),
                         new File(testResourcesDir + "help/expected/HelpDoclet"),
+                        indexFileName,
                         "html", // testIndexFileExtension
                         "html", // testOutputFileExtension
                         "html", // requestedIndexFileExtension
-                        "html"  // requestedOutputFileExtension
+                        "html", // requestedOutputFileExtension
+                        true    // onlyTestIndex
```
magicDGS: Except the BashTestDoclet test, the others should have this set to `false`, no?

src/test/resources/org/broadinstitute/barclay/help/templates/BashTestDoclet/bash-completion.template.ftl
magicDGS: I just realized that this is a test resource, which is fine as it is...but I would like to have a default template for bash-completion which is simple enough to use for any Barclay-derived toolkit...

src/main/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandler.java
jonn-smith: No prob - I noticed that some options were coming out as dashes when I was testing.

src/test/java/org/broadinstitute/barclay/help/DocumentationGenerationIntegrationTest.java
jonn-smith: Absolutely.  Good catch - I'm fixing it now.

src/test/resources/org/broadinstitute/barclay/help/templates/BashTestDoclet/bash-completion.template.ftl
jonn-smith: Yeah - I used this as a test bed for the GATK launch script, but I can make it a little more generic.  In practice, all that's necessary is to change this section at the top of the template (lines 12-20) to be application-specific.  

Right now it assumes there's a central dispatch / launcher script that is used to run the other tools.

cmnbroad: Having generic tab completion capability in Barclay would definitely be a bonus, but I'm not sure what it would take get there. Given that the test in this PR is fairly GATK-specific and isn't really testing Barclay functionality, I'm thinking it should live in GATK. @jonn-smith What do you think about, as a first step, separating out the core Barclay changes for this PR, and moving the rest (the .sh, .ftl, and test) to a PR in the GATK repo ?

src/test/java/org/broadinstitute/barclay/help/BashTestDoclet.java
```diff
@@ -0,0 +1,143 @@
...
+    public static boolean start(RootDoc rootDoc) {
+        try {
+            return new BashTestDoclet().startProcessDocs(rootDoc);
+        } catch (IOException e) {
+            throw new DocException("Exception processing javadoc", e);
+        }
+    }
+
+    @Override
+    protected DocWorkUnit createWorkUnit(
```
jonn-smith: Yup - it's the same.  Pulling out this override.

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
```diff
@@ -350,14 +355,21 @@ private DocumentedFeature getDocumentedFeatureForClass(final Class<?> clazz) {
...
         final Template template = cfg.getTemplate(getIndexTemplateName());
-        final File indexFile = new File(getDestinationDir() + "/index." + getIndexFileExtension());
+
+        final File indexFile = new File(getDestinationDir()
+                            + File.separator
+                            + getIndexBaseFileName()
+                            + "."
+                            + getIndexFileExtension()
+        );
+
```
jonn-smith: Fixed!

src/test/java/org/broadinstitute/barclay/help/BashTestDoclet.java
```diff
@@ -0,0 +1,143 @@
...
+     *             type
+     *             options
+     *             attributes
+     *             required
+     *             minRecValue
+     *             maxRecValue
+     *             minValue
+     *             maxValue
+     *             defaultValue
+     *
```
jonn-smith: No prob.  It'll keep us sane later.

src/test/java/org/broadinstitute/barclay/help/BashTestDoclet.java
```diff
@@ -0,0 +1,143 @@
...
+            } catch (TemplateException e) {
+                throw new DocException("Freemarker Template Exception during documentation index creation", e);
+            }
+
+        } catch (FileNotFoundException e) {
+            throw new RuntimeException("FileNotFoundException processing javadoc template", e);
+        } catch (IOException e) {
+            throw new RuntimeException("IOException processing javadoc template", e);
+        }
+    }
```
jonn-smith: That's a good point.  It's pretty trivial to move the logic into processIndexTemplate - I'll update it.

jonn-smith:
@cmnbroad I think having the test in here makes sense so it can serve as a template for anyone who would want to have tab completion in the future.  I left it in so someone could see how to populate the info for their own purposes.

I'll pull out the GATK-specifics and leave some generic information so that someone can still see how to populate everything.

cmnbroad:
@jonn-smith I like the idea of making this available as an example somehow, but my concern with keeping it as a Barclay test is that its not really testing any feature in Barclay, and we don't have a way to verify that tab completion works, and doesn't regress. Any thoughts on if there is any way to do that, either here or in GATK when we enable it there ?. We already have a bit of a maintenance issue in Barclay due to the coarse-grained nature of the integration tests - whenever we change a feature or add a new test, it results in changes to all of the expected output files, which then have to be upated. Its somewhat manageable, though error prone, for the doc test outputs because they can be loaded into a browser and inspected to see if they look approximately correct. Its harder to do that with the output from tab completion. (We're planning to start using a mocking framework to reduce this problem, but we're not quite there yet, and I don't think it would help for tab completion).

One option would be to move the code to an "Example" package, though it still would be nice to be able to test it somehow. Another option would be to make it into an actual feature in Barclay, but I'm not sure what that would take. Is that a possibility ? Could it be changed so that the caller passes in some argument/s, or template file fragment, that contains the name of the executable and whatever else it needs ?

jonn-smith:
@cmnbroad I've been thinking about testing a lot, and I haven't come up with any good answers on my own.  I agree - all the barclay outputs seem to suffer from similar issues when it comes to test.  

To me the tests here basically function as an example project already - they have code and templates that generate the HTML and now the tab completion as well.  Having the tab completion in there shows that barclay can generate more kinds of files than the documentation you'd expect.  

I don't think I completely follow the example project / barclay feature suggestion.  Would that be akin to generating a barclay jar, then passing that information into barclay manually (as opposed to running the equivalent gradle docgen task)?  If so, sure - it would work, but to make it simple you'd probably want to pass a user-created jar as the argument to barclay, which then is inspected in the normal barclay way to create the file.  IMO this seems to be a little more clunky than leaving the primary interface as gradle.

I'll look around for good closed-loop testing methods for this.  I'm open to suggestion if you (or anyone else) has any ideas.

jonn-smith:
@cmnbroad A cursory look at some testing tools for bash and specifically for tab completion pointed me at this page with a script to test the completion options:

https://brbsix.github.io/2015/11/29/accessing-tab-completion-programmatically-in-bash/

And this page for the Bash Automated Testing Framework:

https://github.com/sstephenson/bats

I think combining the two of them can do the kind of testing that we need in a closed-loop way.

cmnbroad:
The distinction I was making about tab completion being a Barclay "feature" was that it would be like doc, in the sense that if you use Barclay in a project, you can generate doc for "free" by just calling the Barclay doclet with your class list (the equivalent of whats in the GATK build.gradle file). No additional coding is required. GATK does use a custom doclet, and it's own templates, to provide incremental functionality, but there is no Barclay requirement to do that - we could generate doc using Barclay out of the box.

So I'm wondering what it would take to make tab completion like that. As it stands, I think all of the code in this PR (templates & doclet & whatever we wind up with for testing) will also have to be put into GATK as well.

So the best option would be make tab completion like doc, where all we need to put in GATK would be a gradle task to call the Barclay tab completion doclet. Is that an option ? For example, the tab completion code needs to know the name of the app/script ("gatk-launch"), but that could be made into a configurable argument. I'm not really clear on how much more of whats in the templates/.sh is app-specific.

jonn-smith:
Oh, I see what you mean.  Yeah - that should be pretty straight-forward.

For input, you'd need a dispatch script/executable name, along with any options that are high-level options that apply only to that dispatch script/executable (and their relevant information).  If you look at the top of the tab-completion template (bash-completion.template.ftl lines 10-24), there's a section that has caller script options. That's all the information you'd need for application-specific settings.

src/main/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandler.java
```diff
@@ -400,6 +400,11 @@ private void processNamedArgument(
                 argBindings.put("minRecValue", "NA");
                 argBindings.put("maxRecValue", "NA");
             }
+
+            // Add in the number of times you can specify it:
+            argBindings.put("minElements", argDef.field.getAnnotation(Argument.class).minElements());
```
magicDGS: Can you open a different PR for this change? It seems independent from the bash-completion and it will be useful for the normal doclet and this PR will take longer...

src/main/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandler.java
magicDGS: Maybe open an independent PR for this too...I would like to have this in as soon as possible because it's causing me downstream probles...

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
```diff
@@ -488,7 +552,7 @@ protected void processWorkUnitTemplate(
         try {
             // Merge data-model with template
             Template template = cfg.getTemplate(workUnit.getTemplateName());
-            File outputPath = new File(getDestinationDir() + "/" + workUnit.getTargetFileName());
+            File outputPath = new File(getDestinationDir() + File.separator + workUnit.getTargetFileName());
```
magicDGS: Why do not use here just `new File(getDestinationDir(), workUnit.getTargetFileName())`, which use the default file separator?

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
```diff
@@ -510,7 +574,7 @@ protected void processWorkUnitTemplate(
         );
 
         // Convert object to JSON and write JSON entry to file
-        File outputPathForJSON = new File(getDestinationDir() + "/" + workUnit.getJSONFileName());
+        File outputPathForJSON = new File(getDestinationDir() + File.separator + workUnit.getJSONFileName());
```
magicDGS: The same as before, using the constructor for `File`and `String`

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
```diff
@@ -127,26 +146,43 @@ else if (!settingsDir.isDirectory())
      * methods defined here to allow default handling of builtin options.
      *
      * @param options Options to parse.
+     * @return True if {@code options} was parsed, False otherwise.
      */
-    protected void parseOption(final String[] options) {
-        if (options[0].equals(SETTINGS_DIR_OPTION))
+    protected boolean parseOption(final String[] options) {
+
+        boolean hasParsedOption = false;
```
magicDGS: I think that it will be cleaner if this variable is removed, each if/else-if clause is returning true and at the end it returns false. But it is just a matter of taste...

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
```diff
@@ -270,6 +306,11 @@ private void processDocs(final RootDoc rootDoc) {
     public String getIndexTemplateName() { return "generic.index.template.html"; }
 
     /**
+     * @return The base filename for the index file associated with this doclet.
+     */
+    public String getIndexBaseFileName() { return "index"; }
```
magicDGS: I was looking for a chage like this one! Thanks!

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
```diff
@@ -371,14 +431,18 @@ private DocumentedFeature getDocumentedFeatureForClass(final Class<?> clazz) {
      * @param groupMaps
      * @throws IOException
      */
-    private void processIndexTemplate(
+    protected void processIndexTemplate(
```
magicDGS: I wanted also this to be open...can you separate this in a different PR to get it in faster? I would like to be able to process several indexes, and this is the only way of doing it (I guess).

src/test/resources/org/broadinstitute/barclay/help/expected/HelpDoclet/org_broadinstitute_barclay_help_TestArgumentContainer.html
```diff
@@ -538,7 +536,7 @@
 		</p>
 		<hr style="border-bottom: dotted 1px #C0C0C0;" />
 		<h3><a name="--testPlugin">--testPlugin </a>
-			 / <small>-</small>
+			 / <small>NA</small>
```
magicDGS: If the fix for this is in a different PR, it will be nice to change the template to handle NAs properly and remove them from the output.

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
```diff
@@ -0,0 +1,356 @@
...
+
+import java.io.*;
+import java.util.*;
+
+/**
+ * For testing of help documentation generation.
+ */
+public class BashTabCompletionDoclet extends HelpDoclet {
+
+    // Barclay BashTabCompletionDoclet custom Command-line Arguments:
```
magicDGS: Can you add a comment to each of the options and/or variables saying what they are meant for? I have little experience with bash-completion and it will be useful for me to understand the code better. Maybe this could be added to the class javadoc to allow developers to know the options available and what are they use for...

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
```diff
@@ -0,0 +1,356 @@
...
+        } catch (TemplateException e) {
+            throw new DocException("Freemarker Template Exception during documentation index creation", e);
+        }
+    }
+
+    /**
+     * @return the name of the index template to be used for this doclet
+     */
+    @Override
+    public String getIndexTemplateName() { return "bash-completion.template.ftl"; }
```
magicDGS: Nice that the extension is `flt` to show that it is a template. I think that we should do this for the other templates in Barclay...what do you think, @cmnbroad?

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
```diff
@@ -0,0 +1,356 @@
...
+
+    /**
+     * @return The base filename for the index file associated with this doclet.
+     */
+    @Override
+    public String getIndexBaseFileName() { return callerScriptName + "-completion"; }
```
magicDGS: There is two things that I do not fully understand about this, because I am not familiar with bash-completion:

1. If I want to use an uber-jar directly (through `java -jar Toolkit.jar`), is there any possibility to use bash-completion in that case? If it could be done, how this will affect the basename?
2. If the caller script is just a bash wrapper of `java -jar ToolKit.jar` named `toolkit.sh`, should the file be named `toolkit.sh-completion` or `toolkit-completion`?

Sorry for my ignorance, but I would like to know more about this and I think that it is a good opportunity to ask.

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
jonn-smith: Good point - fixed!

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
jonn-smith: Agreed

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
jonn-smith: That's a good idea.  I'll add them in.  I put the comments for the options in the resulting bash as well for review.

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
jonn-smith: 1 - For this case I've been trying to find something to do completion, but because of how bash keys off of the first argument, I'm not sure this will work without hooking into the completion for the `java` invocation itself.  If I can find a way to plug into it, then it should not be very hard to make this work.

2 - I added in the "-completion" notation so that it would be clear from the filename that the file is for tab completion.  To be clear, the name of the resulting script doesn't actually matter for this functionality.  To enable completion you source the script and your shell then has the information it needs to complete the options.

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
jonn-smith: I've always thought that for the most part it's a little less complex to have a single return from a function.

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
magicDGS: I see. Thank you very much for your answer, @jonn-smith!

src/main/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandler.java
jonn-smith: Sure - I'll open another PR with the small bugfixes / changes.

jonn-smith: They're rolled up into #88

magicDGS: Thanks!

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
```diff
@@ -95,6 +105,17 @@ public static boolean start(final RootDoc rootDoc) throws IOException {
...
+         if (!settingsDir.exists()) {
+             logger.error(SETTINGS_DIR_OPTION + " : " + settingsDir.getPath() + " does not exist - will fallback on classpath: " + DEFAULT_SETTINGS_CLASSPATH);
+             settingsDirIsValid = false;
+         }
+         else if (!settingsDir.isDirectory()) {
+             logger.error(SETTINGS_DIR_OPTION + " : " + settingsDir.getPath() + " is not a directory - will fallback on classpath: " + DEFAULT_SETTINGS_CLASSPATH);
+             settingsDirIsValid = false;
+         }
+     }
+
```
cmnbroad: I think we should find a more explicit way to signal the use of the classpath falback; maybe add an explicit argument that says to use the default, and then make it mutually exclusive with settings-dir (throw if they're both provided).

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
cmnbroad: I'm going to agree with both of you...

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
```diff
@@ -127,26 +146,43 @@ else if (!settingsDir.isDirectory())
...
+    protected boolean parseOption(final String[] options) {
+
+        boolean hasParsedOption = false;
+
+        if (options[0].equals(SETTINGS_DIR_OPTION)) {
             settingsDir = new File(options[1]);
-        if (options[0].equals(DESTINATION_DIR_OPTION))
+            hasParsedOption = true;
+        }
+        else if (options[0].equals(DESTINATION_DIR_OPTION)) {
```
cmnbroad: thanks for fixing these ifs - I've been wanting to do that for a while.

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
```diff
@@ -299,9 +340,28 @@ private void emitOutputFromTemplates (
...
+            // Only add the settings directory if it's valid:
+            if ( settingsDirIsValid ) {
+                templateLoaderList.add( new FileTemplateLoader(new File(settingsDir.getPath())) );
+            }
+            templateLoaderList.add( new ClassTemplateLoader(getClass(), DEFAULT_SETTINGS_CLASSPATH) );
+
+            // Set the order in which to load the templates
+            // Here, loading order is the order in which the loaders are specified in the array.
+            // That is, we load from file first (if it's valid), then fall back to class path
+            MultiTemplateLoader multiTemplateLoader =
```
cmnbroad: multiTemplateLoader and templateLoaderList can both be final.

cmnbroad: I'm not sure if we can simplify this template loader code if we go with my suggestion about being explicit about using the fallback, but if we can it would be good.

src/test/java/org/broadinstitute/barclay/help/DocumentationGenerationIntegrationTest.java
```diff
@@ -99,36 +227,48 @@ public void testDocGenRoundTrip(
             final Class<?> docletClass,
             final File inputTemplatesFolder,
             final File expectedDir,
+            final String indexFileBaseName,
             final String testIndexFileExtension,
             final String testOutputFileExtension,
             final String requestedIndexFileExtension,
-            final String requestedOutputFileExtension
+            final String requestedOutputFileExtension,
+            final String[] extraJavadocArgs,
```
cmnbroad: can we call these customDocletArgs ?

src/test/java/org/broadinstitute/barclay/help/DocumentationGenerationIntegrationTest.java
```diff
@@ -68,29 +78,147 @@
...
+                        indexFileName,
+                        "html", // testIndexFileExtension
+                        "html", // testOutputFileExtension
+                        "html", // requestedIndexFileExtension
+                        "html", // requestedOutputFileExtension
+                        new String[] {}, // extraJavadocArgs
+                        false    // onlyTestIndex
+                },
+                // custom bash doclet and templates
+                {BashTabCompletionDoclet.class,
```
cmnbroad: Optional if you don't have the time now, but I'd like to see the tab completion tests in a separate file. We'd have to share the common setup code somehow. If not now, I'd probably want to do it at some point.

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
cmnbroad: It would also be good to document which, if any, are optional (currently I think its all of them, though I think the launch script name should be required).

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
```diff
@@ -0,0 +1,533 @@
...
+            final DocWorkUnit workUnit,
+            final List<Map<String, String>> indexByGroupMaps,
+            final List<Map<String, String>> featureMaps) {
+
+            // For the Bash Test Doclet, this is a no-op.
+            // We only care about the index file.
+    }
+
+    /**
+     * The Index file in the Bash Test Doclet is what generates the actual tab-completion script.
```
cmnbroad: remove "Test"

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
```diff
@@ -0,0 +1,533 @@
...
+     * @throws IOException
+     */
+    @Override
+    protected void processIndexTemplate(
+            final Configuration cfg,
+            final List<DocWorkUnit> workUnitList,
+            final List<Map<String, String>> groupMaps
+    ) throws IOException {
+        // Create a root map for all the work units so we can access all the info we need:
+        Map<String, Object> propertiesMap = new HashMap<>();
```
cmnbroad: I think all of these Maps can be final.

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
```diff
@@ -0,0 +1,533 @@
...
+import com.sun.javadoc.RootDoc;
+import freemarker.template.Configuration;
+import freemarker.template.Template;
+import freemarker.template.TemplateException;
+
+import java.io.*;
+import java.util.*;
+
+/**
+ * For testing of help documentation generation.
```
jonn-smith: Fixed!

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
```diff
@@ -0,0 +1,533 @@
...
+    final private static String CALLER_SCRIPT_POSTFIX_LEGAL_ARGS          = "-caller-post-legal-args";
+    final private static String CALLER_SCRIPT_POSTFIX_ARG_VALUE_TYPES     = "-caller-post-arg-val-types";
+    final private static String CALLER_SCRIPT_POSTFIX_MUTEX_ARGS          = "-caller-post-mutex-args";
+    final private static String CALLER_SCRIPT_POSTFIX_ALIAS_ARGS          = "-caller-post-alias-args";
+    final private static String CALLER_SCRIPT_POSTFIX_ARG_MIN_OCCURRENCES = "-caller-post-arg-min-occurs";
+    final private static String CALLER_SCRIPT_POSTFIX_ARG_MAX_OCCURRENCES = "-caller-post-arg-max-occurs";
+
+    // =============================================
+
+    // Variables that are set on the command line when running javadoc:
```
jonn-smith: All the command-line arguments are optional - I did that so the project could be run and instantly generate documentation without many setup woes.  I agree that the name probably should be provided, but it has a default name that it generates for the script which is somewhat descriptive ("launchScript-completion.sh").

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
```diff
@@ -0,0 +1,533 @@
...
+
+    /**
+     * Name of the executable / wrapper script that will actually invoke the java process.
+     * This wrapper script would call into the JAR and tell it which class to run.
+     */
+    private String callerScriptName = indexDefaultBaseFileName;
```
jonn-smith: Good point on the comment - added one in at the class-level.

I like the idea of being able to just use this out of the box without setting anything up.  @cmnbroad How would you feel about changing `launchScript` to be more descriptive and leaving the argument optional?

To your last point, I'm not sure what you mean.  Once Picard is incorporated into the main GATK project, it should be as simple as re-running the tab completion doclet to create a new tab-completion script.  I don't think much (if anything) would need to change mechanically to make it create that file.

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
```diff
@@ -0,0 +1,533 @@
...
+     * Types of the arguments that the executable / wrapper script is expecting before any Java class names / tools.
+     * The order of these space-delimited types should correspond to the contents of {@link #callerScriptPrefixLegalArgs}
+     *
+     * This is expected to be a space-delimited string of types.
+     *
+     * Currently accepted type values are the following (not case-sensitive):
+     *
+     *      {@code file}
+     *      {@code folder}
+     *      {@code directory}
```
jonn-smith: folder and directory are aliases for the same functionality.  I included this so a user could specify either depending on what they're used to.

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
```diff
@@ -0,0 +1,533 @@
...
+     *      {@code int}
+     *      {@code long}
```
jonn-smith: Similar to folder/directory, int/long and double/float resolve as the same thing (it's in the `_getCompletionWordList` function in the template).  This is here to make it easier for built-in types to complete in java.

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
```diff
@@ -0,0 +1,533 @@
...
+     *      {@code double}
+     *      {@code float}
+     *      {@code null}
```
jonn-smith: Exactly.  If an argument takes no additional specifications (i.e. a flag), then you would set it's type to null.

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
magicDGS: I agree with @cmnbroad about making required the name of the CALLER_SCRIPT_NAME. Even if it works out of box, that is an important parameter that is project-dependent and should be configured: I can imagine situations where many projects depends on Barclay and generate bash-completion with the same name just because they aren't forced to think about that.

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
jonn-smith: I added a comment to make this more clear.

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
```diff
@@ -0,0 +1,533 @@
...
+
+        for (String opt : sentence.split(sep) ) {
+            stringBuilder.append('"');
+            stringBuilder.append(opt);
+            stringBuilder.append('"');
+            stringBuilder.append(sep);
+        }
+
+        // Remove the trailing separator:
+        stringBuilder.deleteCharAt(stringBuilder.length()-1);
```
jonn-smith: Cool.  Fixed!

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
```diff
@@ -0,0 +1,533 @@
...
+        {
+            return 2;
+        }
+        else {
+            return HelpDoclet.optionLength(option);
+        }
+    }
+
+    @Override
+    protected boolean parseOption(String[] option) {
```
jonn-smith: Fixed!

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
```diff
@@ -0,0 +1,533 @@
...
+        boolean hasParsedOption = super.parseOption(option);
+
+        if ( !hasParsedOption ) {
+
+            // Do the stuff that BashTabCompletionDoclet needs:
+            // (Yes this is inefficient because the parent class cycles through input args too.)
+            if (option[0].equals(CALLER_SCRIPT_NAME)) {
+
+                // Remove the last period and anything after it:
+                int lastDotIndex = option[1].lastIndexOf('.');
```
jonn-smith: Fixed!

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
magicDGS: What will happen if an argument is an input file but the `@Argument` annotation is on a `String`? I'm asking this because, som GATK's inputs are `String` to convert them to `java.nio.Path` and work with gcs/hdfs. If this parameters are set to null here, how bash-completion will treat them?

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
jonn-smith: Removed the .template part of the name.

Replaced "Test" in comment with "Completion"

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
```diff
@@ -0,0 +1,533 @@
...
+     *             exclusiveOf
+     *             type
+     *             options
+     *             attributes
+     *             required
+     *             minRecValue
+     *             maxRecValue
+     *             minValue
+     *             maxValue
+     *             defaultValue
```
jonn-smith: Cool.  Added in minElements and maxElements.

This is slightly different in that it also has the callerScriptOptions map, but yeah - the tools map is what is generated by DefaultDocWorkUnitHandler.  @cmnbroad do you want me to move this into there?  If so, I can append it to the end of the comment on processWorkUnit.

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
jonn-smith: Fixed!

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
jonn-smith: No prob.

src/test/java/org/broadinstitute/barclay/help/DocumentationGenerationIntegrationTest.java
jonn-smith: Sure.  Fixed!

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
jonn-smith: I changed the logic as suggested.  Now the user must specify either the settingsDir or to use the default.

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
jonn-smith: Fixed finals.

Yes - now that there is only one possible template I will revert it back to a simple loader.

src/test/java/org/broadinstitute/barclay/help/DocumentationGenerationIntegrationTest.java
jonn-smith: That's a good idea, but I'd like to get this in for now.  Can we open a new issue for this?

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
jonn-smith: Sounds good.  I'll update it to require the script name.

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
```diff
@@ -0,0 +1,533 @@
...
+     *
+     * This syntax is used to pass this information to directly to the bash completion script.
+     *
+     * Example: {@code "--help --info --list --inputFile --outFolder --memSize --multiplier"}
+     */
+    private String callerScriptPrefixLegalArgs       = "";
+
+    /**
+     * Types of the arguments that the executable / wrapper script is expecting before any Java class names / tools.
+     * The order of these space-delimited types should correspond to the contents of {@link #callerScriptPrefixLegalArgs}
```
jonn-smith: Currently there's no validation for this.  We could collapse them, but the problem would persist for the other corresponding caller script args.

src/main/java/org/broadinstitute/barclay/help/BashTabCompletionDoclet.java
jonn-smith: If an argument is given and specified as a string it will not tab complete - the script will know that it should be there, but won't know how to help you complete that option.

In the event of null, the parameter will be treated as a flag with no option to complete.  That is, if you specify an option with a null value here, then hit space and double tab again, the results will be the remaining arguments for the tool (and will not be an option to complete, such as a file or folder).

cmnbroad:
Looks good @jonn-smith. There are a couple of conflicts - can you either resolve them in github or else just rebase. Then we can merge.

jonn-smith:
Sounds good - I'll take a look.

PR #88, opened 2017-07-14
Several small bugfixes and updates.

jonn-smith:
Pulled out some small changes in a larger  PR ( #82 )  into this one for ease of merging.

Added minElements and maxElements to freemarker objects (they were
    present in the data, but not being pushed into freemarker).

Fixed display names not to show up with '-' as values when they are
empty or null.

ProcessIndexTemplate is now protected.

In the default templates, arguments that have no synonyms no longer
render synonyms at all (as opposed to showing up as NA).

### DIFF
--- a/src/main/resources/org/broadinstitute/barclay/helpTemplates/generic.template.html
+++ b/src/main/resources/org/broadinstitute/barclay/helpTemplates/generic.template.html
@@ -31,7 +31,7 @@
 	<#macro argumentDetails arg>
 		<hr style="border-bottom: dotted 1px #C0C0C0;" />
 		<h3><a name="${arg.name}">${arg.name} </a>
-			<#if arg.synonyms??> / <small>${arg.synonyms}</small></#if>
+			<#if arg.synonyms??><#if arg.synonyms != "NA"> / <small>${arg.synonyms}</small></#if></#if>
 		</h3>
 		<p class="args">
 			<b>${arg.summary}</b><br />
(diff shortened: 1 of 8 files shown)


src/main/resources/org/broadinstitute/barclay/helpTemplates/generic.template.html
```diff
@@ -31,7 +31,7 @@
 	<#macro argumentDetails arg>
 		<hr style="border-bottom: dotted 1px #C0C0C0;" />
 		<h3><a name="${arg.name}">${arg.name} </a>
-			<#if arg.synonyms??> / <small>${arg.synonyms}</small></#if>
+			<#if arg.synonyms??><#if arg.synonyms != "NA"> / <small>${arg.synonyms}</small></#if></#if>
```
cmnbroad: Since we're changing this line anyway, can you remove the outer "#<if arg.synonyms??>". I think it probably is and always was unnecessary.

src/test/resources/org/broadinstitute/barclay/help/templates/TestDoclet/generic.template.html
```diff
@@ -31,7 +31,7 @@
 	<#macro argumentDetails arg>
 		<hr style="border-bottom: dotted 1px #C0C0C0;" />
 		<h3><a name="${arg.name}">${arg.name} </a>
-			<#if arg.synonyms??> / <small>${arg.synonyms}</small></#if>
+			<#if arg.synonyms??><#if arg.synonyms != "NA"> / <small>${arg.synonyms}</small></#if></#if>
```
cmnbroad: Same thing here.

src/main/resources/org/broadinstitute/barclay/helpTemplates/generic.template.html
jonn-smith: Sounds good.  Done!

src/test/resources/org/broadinstitute/barclay/help/templates/TestDoclet/generic.template.html
jonn-smith: Sounds good.  Done!

PR #89, opened 2017-07-21
Rename help templates to .ftl

magicDGS:
Minor change to be more explicit about files being a FreeMarker template (.ftl extension).


src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java
```diff
@@ -267,7 +267,7 @@ private void processDocs(final RootDoc rootDoc) {
     /**
      * @return the name of the index template to be used for this doclet
      */
-    public String getIndexTemplateName() { return "generic.index.template.html"; }
+    public String getIndexTemplateName() { return "generic.index.ftl"; }
```
cmnbroad: Same here.

src/main/resources/org/broadinstitute/barclay/helpTemplates/generic.ftl
```diff
@@ -4,7 +4,7 @@
 
 <div class='row-fluid' id="top">
 
-	<#include "common.html"/>
+	<#include "common.ftl"/>
```
cmnbroad: I think common.html should retain it's current name since its just html and isn't really a template. It's name will never be transformed by the doc process, so the end result would be a .html that includes a .ftl if we do this.

src/main/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandler.java
```diff
@@ -24,7 +24,7 @@
     final protected static Logger logger = LogManager.getLogger(DefaultDocWorkUnitHandler.class);
 
     private static final String NAME_FOR_POSITIONAL_ARGS = "[NA - Positional]";
-    private static final String DEFAULT_FREEMARKER_TEMPLATE_NAME = "generic.template.html";
+    private static final String DEFAULT_FREEMARKER_TEMPLATE_NAME = "generic.ftl";
```
cmnbroad: I think it would be useful to retain the ".html" in the file name to indicate what kind of template these are, maybe "generic.html.ftl".

src/main/resources/org/broadinstitute/barclay/helpTemplates/generic.ftl
magicDGS: I think that you are referring to the top php line in the generic template, because this is an `#include` for FreeMaker and the `common.ftl` a set of macros used in both the generic and index templates.

src/main/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandler.java
magicDGS: Strongly agree with this one: `generic.html.ftl` sounds much better!

Done.

src/main/java/org/broadinstitute/barclay/help/HelpDoclet.java

magicDGS:
Back to you @cmnbroad - I've already done the change of the names in the test templates too, but I added now the html.flt extension. And related to the common template, I think that it is just a set of macros and that's why I also called it ftl. Maybe we should rename to `common.macros.html.ftl`?

src/main/resources/org/broadinstitute/barclay/helpTemplates/generic.ftl
cmnbroad: Ah - you're right. It *is* a template. Thx.

cmnbroad:
Thanks again!

magicDGS:
Thank you, @cmnbroad!

Issue #94, opened 2017-09-21
Change the list file argument expansion extension from ".list" to ".args"


cmnbroad:
".list" is commonly used for Picard interval lists, which cannot be subject to Barclay auto-expansion. So we should change the extension to ".args". See https://github.com/broadinstitute/gatk/issues/3555.

PR #95, opened 2017-09-22
Change the list file argument expansion extension and add a new option

magicDGS:
* First commit solves #94 
* Second commit adds an option to switch off the list parsing.

### DIFF
--- a/src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
+++ b/src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
@@ -717,7 +717,8 @@ public final class CommandLineArgumentParser implements CommandLineParser {
     private List<String> expandListFile(final List<String> originalValues) {
         List<String> expandedValues = new ArrayList<>(originalValues.size());
         for (String stringValue: originalValues) {
-            if (stringValue.endsWith(COLLECTION_LIST_FILE_EXTENSION)) {
+            if (!parserOptions.contains(CommandLineParserOptions.DO_NOT_EXPAND_COLLECTION_LIST_FILE)
+                    && stringValue.endsWith(COLLECTION_LIST_FILE_EXTENSION)) {
                 expandedValues.addAll(loadCollectionListFile(stringValue));
             }
             else {
--- a/src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
+++ b/src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
@@ -62,8 +62,10 @@ public final class CommandLineArgumentParser implements CommandLineParser {
     public static final String COMMENT = "#";
     public static final String POSITIONAL_ARGUMENTS_NAME = "Positional Argument";
 
-    // Extension for collection argument list files
-    private static final String COLLECTION_LIST_FILE_EXTENSION = ".list";
+    /**
+     * Extension for collection argument list files.
+     */
+    public static final String COLLECTION_LIST_FILE_EXTENSION = ".args";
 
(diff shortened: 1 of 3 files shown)


magicDGS:
@cmnbroad - do you want to review this one?

lbergelson:
If we're going to do this maybe we should just go all the way and let you specify a list of extension that will be expanded this way.  Default to `.args` but allow people to set empty list for off, or several extensions if they want to support their own set of things?

magicDGS:
What about a `CommandLineParserOptions` interface to have also user-implementations of other options in the parser? The idea is similar to the `java.nio.file.OpenOption`:

* Interface for the `CommandLineParserOption`
* An enum for switch options (e.g. APPEND_TO_COLLECTIONS) that does not require any specific behaviour implementation-dependent.
* Sub-interfaces for other options, as this one. For example, a `ExpandCollectionExtensionOption` with a `Collection<String> getExtensions()` method to return them. For  solving https://github.com/broadinstitute/gatk/issues/3555, GATK may return `.args` (as simple as `() -> Collections.singleton(".args")`), and other toolkits may implement their own.

This can be also useful for https://github.com/broadinstitute/gatk/issues/2596 to specify some syntax checks/enforcements on arguments (e.g., conversion to kebab-case syntax, or just check after a refactor of all arguments is done). What do you think, @lbergelson and @cmnbroad?

cmnbroad:
@magicDGS We'd certainly take a look at a PR if you want to propose something.

magicDGS:
@cmnbroad - I implemented the changes in the next commit to show the idea (slightly modified from the previous description.
@lbergelson - Now there is a way to set none, different, or several extensions.

Back to you!

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -146,14 +151,16 @@ private String getUsagePreamble() {
...
+     *
      * @param callerArguments The object containing the command line arguments to be populated by
      *                        this command line parser.
      */
     public CommandLineArgumentParser(final Object callerArguments) {
         this(
                 callerArguments,
                 Collections.<CommandLinePluginDescriptor<?>>emptyList(),
-                Collections.<CommandLineParserOptions>emptySet()
+                DEFAULT_PARSER_OPTIONS
```
cmnbroad: Since we now want to have more complex options for things like custom expansion file extensions, I think it would be cleaner to switch over to using a configuration object that is passed in by clients. Barclay can have a default implementation, with methods like "List<String> getExpansionFileExtensions()", and custom implementations be provided by the clients.

magicDGS: So do you mean to use a `BarclayConf` interface (or mutable object) to pass to the parser? I think that passing the options as single elements makes clear which behaviour is changing/requested...

If we are going to that direction, I can either make another commit here or open a new PR, whatever you prefer...

cmnbroad: I meant that instead of the CommandLineParserOption enum or interface, Barclay should have a CommandLineParserConfig class, with methods like "List<String> getExpansionFileExtensions()" and "boolean getAppendToCollections()", with default implementations. Any customizations would be implemented by passing a config object to the constructor which is a subclass of that default implementation. Barclay would just query it; hopefully the object itself won't need any mutators.

When we just had options that were toggles, it would have been overkill, but with the need to return a custom list, this is cleaner, and will eliminate the need for type casting, etc. Reusing this PR is fine, or make a new one if its easier - either way.

magicDGS:
@cmnbroad - I added a commit with the `CommandLineParserConfig` implementation that you suggested.

PR #98, opened 2017-10-03
Fix populate arguments from DocumentedFeatures with non-arg constructors

### LINKED ISSUE #97: DocumentedFeatures do not populate arguments
After https://github.com/broadinstitute/barclay/commit/78b17117c264df3a39a1e29a3bf34f323b900eab, even if a documented feature has a non-arg constructor (such as `ReadFilter`), the arguments are not populated into the JSON nor in the property map. This is because they are not annotated with the `CommandLineProgramProperties`.

Maybe it should be enforced by Barclay to have every documented feature with a non-arg constructor, or have a marker for such a argument container.

magicDGS:
Closes #97

### DIFF
--- a/src/main/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandler.java
+++ b/src/main/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandler.java
@@ -191,18 +191,19 @@ public class DefaultDocWorkUnitHandler extends DocWorkUnitHandler {
         // Ex: We may want to document an input or output file format by adding @DocumentedFeature
         // to the format's reader/writer class (i.e. TableReader), and then reference that feature
         // in the extraDocs attribute in a CommandLineProgram that reads/writes that format.
-        if (workUnit.getCommandLineProperties() != null) {
-            try {
-                final Object argumentContainer = workUnit.getClazz().newInstance();
-                if (argumentContainer instanceof CommandLinePluginProvider ) {
-                    pluginDescriptors = ((CommandLinePluginProvider) argumentContainer).getPluginDescriptors();
-                    clp = new CommandLineArgumentParser(
-                            argumentContainer, pluginDescriptors, Collections.emptySet()
-                    );
-                } else {
-                    clp = new CommandLineArgumentParser(argumentContainer);
-                }
-            } catch (IllegalAccessException | InstantiationException e) {
+        try {
+            final Object argumentContainer = workUnit.getClazz().newInstance();
+            if (argumentContainer instanceof CommandLinePluginProvider) {
+                pluginDescriptors = ((CommandLinePluginProvider) argumentContainer).getPluginDescriptors();
+                clp = new CommandLineArgumentParser(
+                        argumentContainer, pluginDescriptors, Collections.emptySet()
+                );
+            } else {
+                clp = new CommandLineArgumentParser(argumentContainer);
+            }
+        } catch (IllegalAccessException | InstantiationException e) {
+            // only throw if the command line properties is set - CLP always require a non-arg constructor
+            if (workUnit.getCommandLineProperties() != null) {
                 throw new RuntimeException(e);
             }
         }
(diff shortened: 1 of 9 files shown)


magicDGS:
@cmnbroad - here its a simple bugfix that benefits from the mocking tests! Can you review? I've already in my code a couple of documented classes failing to populate the args without this fix...

src/main/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandler.java
```diff
@@ -191,19 +191,20 @@ public void processWorkUnit(
...
+            if (argumentContainer instanceof CommandLinePluginProvider) {
+                pluginDescriptors = ((CommandLinePluginProvider) argumentContainer).getPluginDescriptors();
+                clp = new CommandLineArgumentParser(
+                        argumentContainer, pluginDescriptors, Collections.emptySet()
+                );
+            } else {
+                clp = new CommandLineArgumentParser(argumentContainer);
+            }
+        } catch (IllegalAccessException | InstantiationException e) {
+            // only throw if the command line properties is set - CLP always require a non-arg constructor
```
cmnbroad: Change this comment to say "DocumentedFeature does not assume a no-arg constructor unless it is also a CommandLineProgram".

src/test/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandlerUnitTest.java
```diff
@@ -80,6 +81,19 @@ public void testGetGroupSummaryForWorkUnit(final Class<?> docWorkUnitClazz, fina
...
+    private final class ClpWithoutNoArgsConstructor {
+        public ClpWithoutNoArgsConstructor(final int arg) { }
+    }
+
+    @Test(expectedExceptions = RuntimeException.class)
+    public void testProcessWorkUnitWithoutNoArgsConstructorClpThrows() {
+        final DefaultDocWorkUnitHandler handler = getDefaultWorkUnitHandlerInstance();
+        final DocWorkUnit workUnit = createMockWorkUnit(handler, ClpWithoutNoArgsConstructor.class, "", Collections.emptyMap());
+        handler.processWorkUnit(workUnit, Collections.emptyList(), Collections.emptyList());
+    }
```
cmnbroad: Nice! We might as well add the converse as a unit test as well; a DocumentedFeature that doesn't have a no-arg constructor and doesn't have a CommandLineProgram annotation. Then just validate that the resulting workUnit property(s).

src/main/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandler.java
magicDGS: Done.
(Because there is no concept of `CommandLineProgram` yet in Barclay, but only the annotation, I changed the ref to `CommandLineProgramProperties` instead)

src/test/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandlerUnitTest.java
magicDGS: The current mocking framework is not able to process the work unit because it is not mocking the `ClassDoc.fields` and thus it throws a NPE, and even if it is mocking it, the codepaths for the `FieldDocs` will change in #105.

I think that testing the rest of the `processWorkUnit()` should be added in a different PR, because it requires much more work such as testing the different codepaths to populate different properties. I added a disabled test to show the issue...

magicDGS:
Back to you @cmnbroad - there is still more work to do for a fully feature mock-testing framework to enable the kind of tests that you are asking for. Let me know what do you think!

src/test/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandlerUnitTest.java
cmnbroad: Yeah, I see your point. I created https://github.com/broadinstitute/barclay/issues/107 for that. In the meantime lets not hold up these bug fix PRs since they're covered by integration tests.

cmnbroad:
@magicDGS Yeah, lets remove them until we have a solution that allows them to work.

cmnbroad:
Thanks for finding this one @magicDGS!

PR #99, opened 2017-10-05
Legacy parser should throw on optional List argument if not provided.

cmnbroad:
Fixes https://github.com/broadinstitute/picard/issues/950.

### DIFF
--- a/src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
+++ b/src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
@@ -423,13 +423,16 @@ public class LegacyCommandLineArgumentParser implements CommandLineParser {
                 }
                 if (optionDefinition.isCollection) {
                     final Collection<?> c = (Collection<?>) optionDefinition.field.get(optionDefinition.parent);
+                    if (c.size() == 0 && !optionDefinition.optional) {
+                        messageStream.print("ERROR: Option '" + fullName + "' is required");
+                        return false;
+                    }
                     if (c.size() < optionDefinition.minElements) {
                         messageStream.println("ERROR: Option '" + fullName + "' must be specified at least " +
                                 optionDefinition.minElements + " times.");
                         return false;
                     }
-                } else if (!optionDefinition.optional && !optionDefinition.hasBeenSet &&
-                        !optionDefinition.hasBeenSetFromParent && mutextOptionNames.length() == 0) {
+                } else if (!optionDefinition.optional && !optionDefinition.hasBeenSet && mutextOptionNames.length() == 0) {
                     messageStream.print("ERROR: Option '" + fullName + "' is required");
                     if (optionDefinition.mutuallyExclusive.isEmpty()) {
                         messageStream.println(".");
@@ -968,7 +971,6 @@ public class LegacyCommandLineArgumentParser implements CommandLineParser {
         final boolean isCommon;
         boolean hasBeenSet = false;
         boolean hasBeenSetFromOptionsFile = false;
-        boolean hasBeenSetFromParent = false;
         final Set<String> mutuallyExclusive;
 
         private OptionDefinition(final Field field, final Object parent,final String name, final String shortName, final String doc,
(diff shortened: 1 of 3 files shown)


nh13:
@cmnbroad can the optional annotation just be removed?

src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
```diff
@@ -423,13 +423,16 @@ private boolean checkNumArguments() {
                 }
                 if (optionDefinition.isCollection) {
                     final Collection<?> c = (Collection<?>) optionDefinition.field.get(optionDefinition.parent);
+                    if (c.size() == 0 && !optionDefinition.optional) {
+                        messageStream.print("ERROR: Option '" + fullName + "' is required");
+                        return false;
```
nh13: code not covered by tests

src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
```diff
@@ -423,13 +423,16 @@ private boolean checkNumArguments() {
                 }
                 if (optionDefinition.isCollection) {
                     final Collection<?> c = (Collection<?>) optionDefinition.field.get(optionDefinition.parent);
+                    if (c.size() == 0 && !optionDefinition.optional) {
+                        messageStream.print("ERROR: Option '" + fullName + "' is required");
```
nh13: instead of `c.size() == 0`, use `c.isEmpty()`

src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
```diff
@@ -423,13 +423,16 @@ private boolean checkNumArguments() {
...
+                        return false;
+                    }
                     if (c.size() < optionDefinition.minElements) {
                         messageStream.println("ERROR: Option '" + fullName + "' must be specified at least " +
                                 optionDefinition.minElements + " times.");
                         return false;
                     }
-                } else if (!optionDefinition.optional && !optionDefinition.hasBeenSet &&
-                        !optionDefinition.hasBeenSetFromParent && mutextOptionNames.length() == 0) {
+                } else if (!optionDefinition.optional && !optionDefinition.hasBeenSet && mutextOptionNames.length() == 0) {
```
nh13: `length()` -> `isEmpty()`

yfarjoun:
@nh13 and make everything required? I don't understand the suggestion.

nh13:
For non-collections, it's required if no value is set in the class.  For collections, you use min/max elements anyhow.  For options, you it's always optional (duh).

nh13:
@yfarjoun see a similar implementation that does away the need for the `optional` annotation: https://github.com/fulcrumgenomics/sopt/blob/master/src/main/java/com/fulcrumgenomics/sopt/cmdline/ArgAnnotation.java

cmnbroad:
There are still a couple of problems with these changes - pls hold off reviewing until its ready (I made the PR a bit prematurely - I didn't think it would get reviewed so fast!).

src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
cmnbroad: Fixed.

src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
cmnbroad: Fixed.

src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
cmnbroad: mutextOptionNames isn't a collection, its a StringBuilder. No isEmpty on that one.

cmnbroad:
@nh13 We could get rid of the annotation, but then we'd have to propagate those changes to two parsers and all of the downstream projects....

nh13:
@cmnbroad too bad we can't get rid of it, since it seems redundant.  Your call, thanks for thinking about it.

magicDGS:
@nh13 - I think that the `optional` field cannot be removed at the moment also because of https://github.com/broadinstitute/barclay/issues/23

PR #102, opened 2017-10-11
Throw CommandLineParserInternalException for immutable collections

### LINKED ISSUE #100: Barclay could throw a more helpful error message when trying to populate an immutable list argument field.
I noticed that I get a java.lang.UnsupportedOperationException from CommandLineArgumentParser when I have an argument using an immutable list like from `Collections.emptyList().` This throws the following unhelpful stack-trace segment:
```
Caused by: java.lang.UnsupportedOperationException
	at java.util.AbstractList.add(AbstractList.java:148)
	at java.util.AbstractList.add(AbstractList.java:108)
	at org.broadinstitute.barclay.argparser.CommandLineArgumentParser.setArgument(CommandLineArgumentParser.java:697)
	at org.broadinstitute.barclay.argparser.CommandLineArgumentParser.parseArguments(CommandLineArgumentParser.java:418)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.parseArgs(CommandLineProgram.java:217)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMain(CommandLineProgram.java:191)
	at org.broadinstitute.hellbender.Main.runCommandLineProgram(Main.java:131)
```

Since there is no obvious indication of what field is causing the error, this should either throw a more helpful exception pointing to which argument caused the problem or there should be a check that all the requested argument list fields are mutable.

magicDGS:
Closes #100


src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -690,11 +690,17 @@ private void setArgument(ArgumentDefinition argumentDefinition, List<String> val
             if (argumentDefinition.isCollection) {
                 @SuppressWarnings("rawtypes")
                 final Collection c = (Collection) argumentDefinition.getFieldValue();
-                if (value == null) {
-                    //user specified this arg=null which is interpreted as empty list
-                    c.clear();
-                } else {
-                    c.add(value);
```
cmnbroad: There is at least one other place in this method that calls clear on a collection (line 627). We should scan the code for any others.

src/test/java/org/broadinstitute/barclay/argparser/CollectionArgumentUnitTests.java
```diff
@@ -277,6 +277,17 @@ public void testCollectionThatCannotBeAutoInitialized() {
...
+        @Argument
+        public List<String> LIST = Collections.emptyList();
+    }
+
+    @Test(expectedExceptions = CommandLineException.CommandLineParserInternalException.class)
+    public void testCollectionThatIsImmmutable() {
+        final ImmutableCollectionArguments o = new ImmutableCollectionArguments();
+        new CommandLineArgumentParser(o).parseArguments(System.err, new String[]{"--LIST", "A"});
+    }
+
```
cmnbroad: We need to have tests for all combinations of APPEND_TO_COLLECTIONS, "null", and real values.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
magicDGS: Done. I added a method to catch the exception while applying a consumer (either clear/add, but in the future may be others).

src/test/java/org/broadinstitute/barclay/argparser/CollectionArgumentUnitTests.java
magicDGS: Done. Added data provider and different combinations.

magicDGS:
@cmnbroad - back to you (this should pass the tests now).

magicDGS:
Friendly ping here, @cmnbroad!

Issue #104, opened 2017-10-19
Field level javadoc can be incorrect


cmnbroad:
During documentation generation, Barclay searches for the FieldDoc corresponding to each command line argument in order to propagate the field's javadoc to the output. When it doesn't find the FieldDoc in the hierarchy of the ClassDoc for the argument's containing class (i.e., when the field is contained in an object that is dynamically created by the tool, such as a plugin object, or by subclass implementations for things like optional/required), it searches through all of available ClassDocs, but isn't discriminating enough can can falsely find the wrong one. The only manifestation of this is that the field comments could be wrong (the rest of the attributes are determined by the argDef's field object, not the FieldDoc object).

cmnbroad:
The current code considers a FieldDoc as matching if the field has the same name as the target argument field, but the fix is to reject the FieldDoc unless it has the same fully qualified name.

PR #105, opened 2017-10-20
Locate FieldDocs for arguments by lookup rather than brute force.

cmnbroad:
Fixes https://github.com/broadinstitute/barclay/issues/104. Also, this reduces the number of "Unable to access field backtrace" error messages produced during docgen to 0 (thelarge number of which was causing [this](https://github.com/broadinstitute/gatk/issues/3710)) for the current GATK and Picard doc builds, since FieldDocs are now retrieved directly from javadoc by name rather than by brute force search through all ClassDoc objects. Tested with current source for both Picard and GATK.


magicDGS:
@cmnbroad - I think that this is a good opportunity to start including mocking tests instead of updating only the test files...

magicDGS:
True @cmnbroad - I realized in my own PR...

PR #116, opened 2017-11-29
Add @Experimental annotation, test fixes.

cmnbroad:
This adds the `@Experimental` annotation (mutually exclusive with` @Beta`). It also fixes a couple of other bugs (we weren't adding the beta property to the json output, and some of the test templates weren't respecting the data type of the freemark map entries; more incentive to develop the unit testing framework once we can get to it).

### DIFF
--- a/src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
+++ b/src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
@@ -50,7 +50,8 @@ public final class CommandLineArgumentParser implements CommandLineParser {
     private static final String defaultUsagePreamble = "Usage: program [arguments...]\n";
     private static final String defaultUsagePreambleWithPositionalArguments =
             "Usage: program [arguments...] [positional-arguments...]\n";
-    protected static final String BETA_PREFIX = "\n\n**BETA FEATURE - FOR EVALUATION ONLY**\n\n";
+    protected static final String BETA_PREFIX = "\n\n**BETA FEATURE - WORK IN PROGRESS**\n\n";
+    protected static final String EXPERIMENTAL_PREFIX = "\n\n**EXPERIMENTAL FEATURE - USE AT YOUR OWN RISK**\n\n";
 
     private static final String NULL_STRING = "null";
     public static final String COMMENT = "#";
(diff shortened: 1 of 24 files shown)


src/main/java/org/broadinstitute/barclay/help/DocWorkUnit.java
```diff
@@ -169,6 +172,11 @@ public void processDoc(final List<Map<String, String>> featureMaps, final List<M
      */
     public boolean getBetaFeature() { return betaFeature != null; }
 
+    /**
+     * @return a boolean determining if this documented feature is marked as an experimental feature
+     */
+    public boolean getExperimentalFeature() { return experimentalFeature != null; }
```
magicDGS: Shouldn't this be (and the previous) `isExperimentalFeature()`?

cmnbroad:
@lbergelson can you take a look at this if you have the time?

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -51,6 +51,7 @@
     private static final String defaultUsagePreambleWithPositionalArguments =
             "Usage: program [arguments...] [positional-arguments...]\n";
     protected static final String BETA_PREFIX = "\n\n**BETA FEATURE - FOR EVALUATION ONLY**\n\n";
+    protected static final String EXPERIMENTAL_PREFIX = "\n\n**EXPERIMENTAL FEATURE - FOR EVALUATION ONLY**\n\n";
```
lbergelson: Should there be a difference in these statements?

src/main/java/org/broadinstitute/barclay/argparser/ExperimentalFeature.java
```diff
@@ -0,0 +1,13 @@
+package org.broadinstitute.barclay.argparser;
+
+import java.lang.annotation.*;
+
+/**
+ * Marker interface for features that are experimental and not for production use.
```
lbergelson: Maybe elaborate more here?  Ex: these tools may never become stable and may be removed or changed dramatically.

src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
```diff
@@ -76,6 +73,9 @@
     private static final int OPTION_COLUMN_WIDTH = 30;
     private static final int DESCRIPTION_COLUMN_WIDTH = 90;
 
+    protected static final String BETA_PREFIX = "\n\n**BETA FEATURE - FOR EVALUATION ONLY**\n\n";
```
lbergelson: It's funny that these constants are defined here and in the other parser.

src/main/java/org/broadinstitute/barclay/help/DocWorkUnit.java
lbergelson: that would be a more idiomatic name

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
cmnbroad: Done.

src/main/java/org/broadinstitute/barclay/argparser/ExperimentalFeature.java
cmnbroad: Done.

src/main/java/org/broadinstitute/barclay/help/DocWorkUnit.java
cmnbroad: Done, both changed.

PR #118, opened 2017-12-10
Disable "null..." comment in primitive types

yfarjoun:
- Removed note "This option can be set to 'null' to clear the default value." for primitive types.
- Added a test to show that it works
- Made sure that there's at least one space between arguments and description

### DIFF
--- a/src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
+++ b/src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
@@ -644,7 +644,7 @@ public class LegacyCommandLineArgumentParser implements CommandLineParser {
         }
 
         int numSpaces = OPTION_COLUMN_WIDTH - optionLabel.length();
-        if (optionLabel.length() > OPTION_COLUMN_WIDTH) {
+        if (optionLabel.length() > OPTION_COLUMN_WIDTH - 1) {
             sb.append("\n");
             numSpaces = OPTION_COLUMN_WIDTH;
         }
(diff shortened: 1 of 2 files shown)


src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
```diff
@@ -644,7 +644,7 @@ private void printOptionParamUsage(final StringBuilder sb, final String name, fi
         }
 
         int numSpaces = OPTION_COLUMN_WIDTH - optionLabel.length();
-        if (optionLabel.length() > OPTION_COLUMN_WIDTH) {
+        if (optionLabel.length() > OPTION_COLUMN_WIDTH - 1) {
```
magicDGS: Does this only happen in the legacy parser? I guess that for concordance, it should be modified in the `CommandLineArgumentParser`

src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
```diff
@@ -672,7 +672,7 @@ private String makeOptionDescription(final OptionDefinition optionDefinition) {
             sb.append("Default value: ");
             sb.append(optionDefinition.defaultValue);
             sb.append(". ");
-            if (!optionDefinition.defaultValue.equals("null")) {
+            if (!optionDefinition.defaultValue.equals("null") && !optionDefinition.field.getType().isPrimitive() ) {
```
magicDGS: This line does not appear in the `CommandLineArgumentParser` implementation. Is it better to remove from here too?

yfarjoun:
No as the two parsers deal differently with the input "null" (which I think
is not allowed in the new parser)

yfarjoun:
@cmnbroad when you have a moment...

src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
cmnbroad: Can you make this change in the other parser as well ?

src/main/java/org/broadinstitute/barclay/argparser/LegacyCommandLineArgumentParser.java
cmnbroad: The other parser does allow "null" for optional args. There are some tests there for it, but please make the same change to check for isPrimitive and add the same usage test there as well.

PR #122, opened 2017-12-20
Changed argument parser to handle argument collections on PluginDescriptors

### LINKED ISSUE #121: CommandLineArgumentParser help output doesn't include plugin values
Plugin descriptors have one or more string collection arguments that hold the plugins the user wants to enable  (i.e, the read filter plugin descriptor has an argument for which read filters the user wants to use). The --help output for these arguments should include the list of possible values, which are dynamically discovered by the plugin system.

The [bug](https://github.com/broadinstitute/barclay/blob/master/src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java#L823) is that the code assumes that such arguments are fields immediately within the descriptor class, which is myopic, and incorrect, since these are often embedded in a dynamically created argument collection object).

jamesemery:
I accomplished this by making the PluginDescriptor responsible for knowing which arguments belong to it and passing every argument to the descriptor. 

Fixes #121

### DIFF
--- a/src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
+++ b/src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
@@ -842,19 +842,19 @@ public final class CommandLineArgumentParser implements CommandLineParser {
     private void usageForPluginDescriptorArgumentIfApplicable(final ArgumentDefinition argDef, final StringBuilder sb) {
         if (CommandLineParser.getUnderlyingType(argDef.field).equals(String.class)) {
             for (CommandLinePluginDescriptor<?> descriptor : pluginDescriptors.values()) {
-                // this argument came from a plugin descriptor; delegate to get the list of allowed values
+                // See if this this argument came from a plugin descriptor; delegate to get the list of allowed values if it is
                 final Set<String> allowedValues = descriptor.getAllowedValuesForDescriptorArgument(argDef.getLongName());
-                if (allowedValues == null) {
-                    // Do nothing because the argument doesn't belong to this descriptor
-                } else if (allowedValues.isEmpty()) {
-                    sb.append("Any value allowed");
-                    return;
-                } else {
-                    sb.append("Possible Values: {");
-                    sb.append(String.join(", ", allowedValues.stream().sorted(String::compareToIgnoreCase).collect(Collectors.toList())));
-                    sb.append("}");
+                if (allowedValues != null) {
+                    if (allowedValues.isEmpty()) {
+                        sb.append("Any value allowed");
+                    } else {
+                        sb.append("Possible Values: {");
+                        sb.append(String.join(", ", allowedValues.stream().sorted(String::compareToIgnoreCase).collect(Collectors.toList())));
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -844,18 +839,26 @@ private String makeArgumentDescription(final ArgumentDefinition argumentDefiniti
...
-            sb.append("Any value allowed");
-        } else {
-            allowedValues.sort(String.CASE_INSENSITIVE_ORDER);
-            sb.append("Possible Values: {");
-            sb.append(String.join(", ", allowedValues));
-            sb.append("}");
+    private void usageForPluginDescriptorArgumentIfApplicable(final ArgumentDefinition argDef, final StringBuilder sb) {
+        if (CommandLineParser.getUnderlyingType(argDef.field).equals(String.class)) {
+            for (CommandLinePluginDescriptor<?> descriptor : pluginDescriptors.values()) {
+                // this argument came from a plugin descriptor; delegate to get the list of allowed values
```
cmnbroad: This comment is out of place now (we haven't yet concluded that this arg came from a descriptor). Maybe just prefix with "See if this argument came from..."

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -844,18 +839,26 @@ private String makeArgumentDescription(final ArgumentDefinition argumentDefiniti
...
+                } else if (allowedValues.isEmpty()) {
+                    sb.append("Any value allowed");
+                    return;
+                } else {
+                    sb.append("Possible Values: {");
+                    sb.append(String.join(", ", allowedValues.stream().sorted(String::compareToIgnoreCase).collect(Collectors.toList())));
+                    sb.append("}");
+                    return;
+                }
+            }
```
cmnbroad: Minor nit, but maybe rewrite this with a single return:

```
                if (allowedValues != null) {
                    if (allowedValues.isEmpty()) {
                        sb.append("Any value allowed");
                    } else {
                        sb.append("Possible Values: {");
                        sb.append(String.join(", ", allowedValues.stream().sorted(String::compareToIgnoreCase).collect(Collectors.toList())));
                        sb.append("}");
                    }
                    return;
                }
            }
```

src/test/java/org/broadinstitute/barclay/argparser/CommandLinePluginUnitTest.java
```diff
@@ -164,8 +169,7 @@ public boolean isDependentArgumentAllowed(final Class<?> dependentClass) {
             if (longArgName.equals(testPluginArgumentName)) {
                 return testPlugins.keySet();
             }
-            throw new IllegalArgumentException("Allowed values request for unrecognized string argument: " + longArgName);
-        }
+            return null; }
```
cmnbroad: } on separate line.

jamesemery:
@cmnbroad Responded to your comments

PR #123, opened 2018-01-03
Kebabify special argument collection args.

cmnbroad:


cmnbroad:
@lbergelson Can one of you take a look at this ?

cmnbroad:
@vdauwera maybe ? I'd like to get this in for the barclay release tomorrow. Its pretty small.

magicDGS:
@lbergelson is right, I would like to have a way to provide my own for my own transition to kebab-case, scheduled not soon enough to be able to apply it by Jan 9th. Thank you in advance!

cmnbroad:
@magicDGS I recognize that ideally these would be configurable, an I'd like to see that happen, but we won't be able to do it in time for the initial GATK release. Is your concern around downstream tools based on GATK (which is still in beta, so there is no guaranty such changes won't be made - as you know we're changing all of the args to kebab case for the initial release). Or tools with a direct dependency on Barclay, independent of GATK ? If its the latter, I don't want to do something that breaks your users, but this will be a major version release bump for Barclay (2.0). We need to be able to make such changes when doing major version releases.

magicDGS:
Ok, I can tolerate that... although it is a breaking change, I think that no user of ny downstream software is using the arguments in this PR. Thanks for considering my concerns!

magicDGS:
I was writing all the comments without having access to a proper computer to add a detailed comment on this. I think that it is much simpler to add the `CommandLineParserConfig` implementation of https://github.com/broadinstitute/barclay/pull/95, including the special argument flags (`getHelpFlag`, `getShowHiddenFlag`, `getArgumentsFileFlagp`). As configurable options grow, this will be more flexible by extending the configuration file and overriding the needed methods.

If you have a look to that PR I can include the getters for the flags for having it for release of GATK4 and do not break compatibility in Barclay again when making this configurable...

cmnbroad:
I had to cut the release in time to get into picard/gatk, so this didn't make it in.

magicDGS:
For making it more general, we can add some getters with default values to `CommandLineParserConfig` (see #126)

lbergelson:
@cmnbroad Can we merge this one?  I think Magic's objections are valid but I think changing this an revisiting them later if someone expresses interest again would be a good idea.

PR #125, opened 2018-01-19
Add suppressFileExpansion property to @Argument.

cmnbroad:
accept .list in addition to .args as a file expansion extension, and handle display representation of long command lines.

### DIFF
--- a/src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
+++ b/src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
@@ -57,8 +57,13 @@ public final class CommandLineArgumentParser implements CommandLineParser {
     public static final String COMMENT = "#";
     public static final String POSITIONAL_ARGUMENTS_NAME = "Positional Argument";
 
-    // Extension for collection argument list files
-    static final String COLLECTION_LIST_FILE_EXTENSION = ".args";
+    /**
+     * Extensions recognized for collection argument file expansion.
+     */
+    static final String COLLECTION_LIST_FILE_EXTENSIONS_LIST = ".list";
+    static final String COLLECTION_LIST_FILE_EXTENSIONS_ARGS = ".args";
+    static final Set<String> COLLECTION_LIST_FILE_EXTENSIONS = new HashSet<>(
+            Arrays.asList(COLLECTION_LIST_FILE_EXTENSIONS_LIST, COLLECTION_LIST_FILE_EXTENSIONS_ARGS));
 
     private static final Logger logger = LogManager.getLogger();
 
@@ -630,7 +635,7 @@ public final class CommandLineArgumentParser implements CommandLineParser {
                 final Collection c = (Collection) argumentDefinition.getFieldValue();
                 c.clear();
             }
-            values = expandListFile(values);
+            values = expandListFile(argumentDefinition, values);
         }
 
         for (int i = 0; i < values.size(); i++) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -57,8 +57,13 @@
     public static final String COMMENT = "#";
     public static final String POSITIONAL_ARGUMENTS_NAME = "Positional Argument";
 
-    // Extension for collection argument list files
-    static final String COLLECTION_LIST_FILE_EXTENSION = ".args";
+    /**
+     * Extensions recognized for collection argument file expansion.
+     */
+    static final String COLLECTION_LIST_FILE_EXTENSIONS_LIST = ".list";
+    static final String COLLECTION_LIST_FILE_EXTENSIONS_ARGS = ".args";
```
magicDGS: I think that it might be better to add the class for configuration at this stage (see https://github.com/broadinstitute/barclay/pull/95), because each of this changes is generating a breaking change. I think that the configuration is a breaking change but flexible enough to avoid the change  of the collection file extension here and keep it only in GATK.

I will open a new PR with that solution to have a clean branch instead of an outdated PR that get lost because its purpose was changed.

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -709,24 +714,34 @@ private void setArgument(ArgumentDefinition argumentDefinition, List<String> val
...
-    private List<String> expandListFile(final List<String> originalValues) {
-        List<String> expandedValues = new ArrayList<>(originalValues.size());
-        for (String stringValue: originalValues) {
-            if (stringValue.endsWith(COLLECTION_LIST_FILE_EXTENSION)) {
-                expandedValues.addAll(loadCollectionListFile(stringValue));
-            }
-            else {
-                expandedValues.add(stringValue);
+    private List<String> expandListFile(final ArgumentDefinition argDef, final List<String> originalValues) {
+        if (!argDef.suppressFileExpansion) {
```
magicDGS: I like this change (although I prefer to being able to change the extension in a per-argument basis), but the configuration sounds much better for the change of the extensions itself. Have a look to the clean PR (I will comment with the number once it is here).

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
cmnbroad: Taking #95 would require a major version bump (by semantic versioning, which I'd like to follow as much as possible), which we just did, and which I'd like to avoid until we can bundle any other such changes together. This PR introduces only a minor change in runtime behavior (no source or binary incompatibility), and the runtime behavior change has a pretty easy workaround (to change the file extension being used).

It think we've accumulated enough feature requests that involve customization that we need a even more general mechanism than the config class as in #95. Instead, I think we should take this change as is, and implement the scheme outlined in (https://github.com/broadinstitute/barclay/issues/127) is more sustainable in the long run.

magicDGS: Sounds reasonable. Looking forward to the next version bump with #127 implemented (and maybe we can add the #68 (although it will require to clean up the branch, which is quite outdated).

src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
```diff
@@ -709,24 +714,34 @@ private void setArgument(ArgumentDefinition argumentDefinition, List<String> val
     }
 
     /**
-     * Expand any collection values that are ".list" argument files, and add them
-     * to the list of values for that argument.
+     * Expand any collection value that references a filename that ends in one of the accepted expansion
+     * extensions, and add the contents of the file to the list of values for that argument.
+     * @param argDef the ArgumentDefinition being populated
      * @param originalValues
```
jonn-smith: Add explanation of parameter

jonn-smith:
@cmnbroad This looks fine to me.  One very minor comment.

PR #141, opened 2018-06-26
Add test for String initialized to null.

cmnbroad:
Came up as a question and I realized there is no test coverage for this case.


cmnbroad:
Hm, I don't think there is, so its not possible to set a String argument to the string value "null" from the command line.

Issue #145, opened 2018-10-17
Barclay CLP should reject hybrid posix syntax (-I=somevalue)


cmnbroad:
Most of the time this works because jopt accepts it, however, if `somevalue` contains an embedded ":", the CLP tag parser is triggered. It sees `I=somevalue` as the option name, and complains that there is no value for the option to consume (assuming there are no additional command line args).

The parser should probably uniformly reject the hybrid syntax by rejecting option names that contain an embedded `=`, which will prevent the tag parser from processing the embedded `:` as a tag delimiter.

PR #146, opened 2018-10-18
Reject attempts to use hybrid Barclay/legacy syntax with the Barclay parser.

cmnbroad:
Fixes https://github.com/broadinstitute/barclay/issues/145.


src/main/java/org/broadinstitute/barclay/argparser/TaggedArgumentParser.java
```diff
@@ -142,11 +143,22 @@ private void replaceTaggedOption(
...
                 // the option appears to be tagged, but we're already at the end of the argument list,
                 // and there is no companion value to use
-                throw new CommandLineException("No value found for tagged argument: " + optionString);
+                throw new CommandLineException("No argument value found for tagged argument: " + optionString);
             }
         }
     }
 
+    // Reject attempts to use hybrid Barclay/legacy syntax that contains embedded "=". Most of the time
+    // this works because jopt accepts "-O=value". But if "value" contains what appears to be tagging
```
lbergelson: Why not just use a javadoc comment?  I've never really understood the prohibition on putting javadoc on private things.

cmnbroad: Just due to habit. Fixed.

PR #156, opened 2020-01-10
Fix error reporting for null argument collections.

### LINKED ISSUE #155: Late failure on a argument collection field with no initialization value (therefore null) v
When I forget to give a value to a argument collection tool argument I get a NPE stacktrace that makes difficult to diagnose and correct the problem: 

```
java.lang.NullPointerException
	at org.broadinstitute.barclay.argparser.CommandLineArgumentParser.createArgumentDefinitions(CommandLineArgumentParser.java:196)
	at org.broadinstitute.barclay.argparser.CommandLineArgumentParser.createArgumentDefinitions(CommandLineArgumentParser.java:210)
	at org.broadinstitute.barclay.argparser.CommandLineArgumentParser.<init>(CommandLineArgumentParser.java:182)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.getCommandLineParser(CommandLineProgram.java:479)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.parseArgs(CommandLineProgram.java:232)
	at org.broadinstitute.hellbender.cmdline.CommandLineProgram.instanceMain(CommandLineProgram.java:206)
	at org.broadinstitute.hellbender.Main.runCommandLineProgram(Main.java:163)
	at org.broadinstitute.hellbender.Main.mainEntry(Main.java:206)
	at org.broadinstitute.hellbender.Main.main(Main.java:292)
```
This may also affect other arguments that are not argument-collections but I have not tested it. 

The code responsible seems to be in ```CommandLineArgumentParser.java```  around line 218:

```java
if (field.getAnnotation(ArgumentCollection.class) != null) {
                try {
                    field.setAccessible(true);
                    createArgumentDefinitions(field.get(callerArguments), controllingDescriptor);
                } catch (final IllegalAccessException e) {
                    throw new CommandLineException.ShouldNeverReachHereException("should never reach here because we setAccessible(true)", e);
                }
            }
```

The problem is that we don't check the result of ```field.get(callerArguments)``` is ```null``` which eventually results in the exception.

Ideally this value should be checked for null and report it in a nicer exception indicating the name of the field and tool class. 

An alternative is that a null is interpreted as "use the no-argument constructor to create the default argument-collection".

cmnbroad:
Fixes #155.

### DIFF
--- a/src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
+++ b/src/main/java/org/broadinstitute/barclay/argparser/CommandLineArgumentParser.java
@@ -335,7 +335,15 @@ public final class CommandLineArgumentParser implements CommandLineParser {
             else if (argumentCollectionAnnotation != null) {
                 try {
                     field.setAccessible(true);
-                    createArgumentDefinitions(field.get(callerArguments), controllingDescriptor);
+                    final Object fieldInstance = field.get(callerArguments);
+                    if (fieldInstance == null) {
+                        throw new CommandLineException.CommandLineParserInternalException(
+                                String.format(
+                                        "The ArgumentCollection field '%s' in '%s' must have an initial value",
+                                        field.getName(),
+                                        callerArguments.getClass().getName()));
+                    }
+                    createArgumentDefinitions(fieldInstance, controllingDescriptor);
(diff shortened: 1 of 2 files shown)


cmnbroad:
Thanks @lbergelson.

PR #159, opened 2020-03-04
*Allow processPositionalArguments to be overridden in docgen subclasses.

cmnbroad:


src/main/java/org/broadinstitute/barclay/help/DefaultDocWorkUnitHandler.java
```diff
@@ -439,7 +439,7 @@ private FieldDoc getFieldDocForCommandLineArgument(
         return fieldDoc;
     }
 
-    private void processPositionalArguments(
+    protected void processPositionalArguments(
```
lbergelson: actually, this should probably have javadoc if it's protected....

cmnbroad:
thank you Louis!

PR #165, opened 2020-08-03
WDL generation updates

cmnbroad:
Change WorkflowResource to WorkflowInputFile/WorkflowOutputFile
Change RuntimeProperties to WorkflowProperties
Add a WorkflowInput property to control localization “optional”
Change cores default to 2
Separate optional/required companion files
Include optional output companions in outputs
Update summary used for companion resources
Fix companion arg doc (reference source name rather than summary).
Added tests, and fixes for positional doc and companion initial values
Factor out some template property name string constants from my incomplete map wrapper class branch
Suppress excess test output to avoid exceeding log output length on travisFix initial value for all args optional (not quoted correctly)
Expose LONG_OPTION_PREFIX/SHORT_OPTION_PREFIX from TaggedArguments


src/main/java/org/broadinstitute/barclay/help/WDLWorkUnitHandler.java
```diff
@@ -183,71 +115,129 @@ protected void processPositionalArguments(
...
+                }
+            }
+            for (final String companion : workflowOutput.optionalCompanions()) {
+                final Map<String, String> companionMap = createCompanionMapEntry(wdlName, companion);
+                runtimeOutputs.put(companionMap.get(TemplateProperties.ARGUMENT_NAME), wdlType);
+                if (!resourceIsOptional) {
+                    requiredOutputs.put(companionMap.get(TemplateProperties.ARGUMENT_NAME), wdlType);
+                    requiredCompanions.add(companionMap);
+                }
+                optionalCompanions.add(companionMap);
```
jonn-smith, line 227: Do you need to add in the optional companions here even if `resourceIsOptional` is true?  I'm a little confused about why you might potentially add the same `companionMap` twice - once to optional and once to required.

src/main/resources/org/broadinstitute/barclay/helpTemplates/wdlToolTemplate.wdl.ftl
```diff
@@ -132,7 +132,15 @@ task ${name} {
   output {
     <@defineTaskOutputs heading="Task Outputs" outputs=runtimeOutputs/>
   }
- }
+  <#if localizationOptional?size != 0>
+
+  parameter_meta {
+    <#list localizationOptional as optLoc>
```
jonn-smith, line 138: Something to think about - should we move the comments in the header describing each input variable to the parameter meta section since that's what it's for?

src/main/java/org/broadinstitute/barclay/help/WDLWorkUnitHandler.java
cmnbroad: Oh yeah, good catch. It is a bit confusing because the optional and required companions are mutually exclusive, but the optional and required outputs are not (this is purely by convention, because it helps make the template code less complex). I'll add some comments and maybe update the names to reenforce this.

src/main/resources/org/broadinstitute/barclay/helpTemplates/wdlToolTemplate.wdl.ftl
cmnbroad: Hm, yeah, that probably makes sense.

src/main/java/org/broadinstitute/barclay/help/WDLWorkUnitHandler.java
cmnbroad: I fixed the bug here, and also updated many of the names in this method to be a bit more explicit. I also updated some residual references to `WorkflowResource` left over in this file.

src/main/resources/org/broadinstitute/barclay/helpTemplates/wdlToolTemplate.wdl.ftl
cmnbroad: I added a param_meta section to both the workflow and task, which cromwell and womtool both seem to be happy with. I also left the comments at the top since it seems harmless and they're easy to scan there. Let me know if you disagree.

cmnbroad:
Back to you @jonn-smith.

PR #166, opened 2020-08-06
Ability to construct a file from a URI

markjschreiber:
The change allows programs like Picard to accept input arguments that are Strings representing either file paths or file scheme URIs.
For example when Picard is built with this new Barclay as a dependency the following are both possible:

`java -jar /Users/mrschre/dev/picard/build/libs/picard.jar SortVcf -I file:///Users/mrschre/dev/picard/testdata/picard/arrays/input.vcf -O file:///tmp/output.vcf`
as well as 
`java -jar /Users/mrschre/dev/picard/build/libs/picard.jar SortVcf -I /Users/mrschre/dev/picard/testdata/picard/arrays/input.vcf -O` /tmp/output.vcf`

Combinations of file paths and file schema URIs are also possible.

### DIFF
--- a/src/main/java/org/broadinstitute/barclay/argparser/ArgumentDefinition.java
+++ b/src/main/java/org/broadinstitute/barclay/argparser/ArgumentDefinition.java
@@ -278,7 +278,9 @@ public abstract class ArgumentDefinition {
      */
     public static File fileFromFileSchemeURI(final String uri) {
         if (uri == null) throw new IllegalArgumentException("File URI cannot be null");
-        return new File(URI.create(uri));
+        // compensate for applications that don't correctly endode # and ? which may never legally appear in a URI
+        String tolerantUri = uri.replace("#", "%23").replace("?", "%3F");
+        return new File(URI.create(tolerantUri));
     }
 
     /**
--- a/src/main/java/org/broadinstitute/barclay/argparser/ArgumentDefinition.java
+++ b/src/main/java/org/broadinstitute/barclay/argparser/ArgumentDefinition.java
@@ -2,8 +2,10 @@ package org.broadinstitute.barclay.argparser;
 
 import org.broadinstitute.barclay.utils.Utils;
 
+import java.io.File;
 import java.io.PrintStream;
 import java.lang.reflect.*;
+import java.net.URI;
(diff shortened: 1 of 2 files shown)


cmnbroad:
@markjschreiber Thanks for submitting a Barclay pull request. The problem you describe (i.e., Picard tools accepting URIs) is one we'd like to address, but at the application rather, than in Barclay. URIs are percent-encoded, so converting a URI to a string suitable for use with the `File` constructor requires the string to be decoded; just stripping off the scheme prefix doesn't do that, and can result in confusing downstream failures for pathnames that contain embedded special characters. We've actually had quite a few issues in Picard, GATK, and htsjdk because there are many code paths that use exactly that same conversion pattern (see https://github.com/samtools/htsjdk/pull/1487). Also, a Barclay solution would also need to be duplicated in the legacy parser (in order to ensure parity between the parsers).

There is a pull request in htsjdk https://github.com/samtools/htsjdk/pull/1496 that contains a class for use in place of `File` inputs that addresses these and other issues (we already use it in many GATK tools). Once its available to Picard, I hope we can start using it there to enable Picard to accept a wider range of input and output URIs such as google cloud storage, hdfs, etc.

markjschreiber:
Good point about the URL encoding. I can certainly URL decode `file://` URIs. Like you say a fix in Barclay isn't the best long term fix, mainly I was hoping this "bandaid" would be useful to others until Picard and other tools are fixed/ improved. Let me know if you think this should be merged (with the encoding fix), made into a separate tag that could be used for people wanting a quick fix to Picard, or just abandoned.

markjschreiber:
I have included a test that demonstrates that encoded URIs are correctly handled. I could make the method more tolerant by allowing it to handle `#` and `?` characters that should never appear in a file URI (because a file path can't have a fragment or a query) but which sometimes do if the producing program doesn't properly encode the URI.

cmnbroad:
@markjschreiber Sorry for the delay in response. As I mentioned in my last comment, I don't think Barclay is the right place to resolve this issue. Once the next version of hsjdk is released, it will provide a path forward for Picard (though it still won't handle the "tolerant"/malformed URIs you supported here - I'm not sure why thats necessary though).

PR #160, opened 2020-03-04
*Generate doc for options that are collections of enums.

cmnbroad:
This adds support for including documentation for argruments that are collections of an enum type.

The actual code change is pretty small, but there are multiple expected output files updated in this PR because all of the tests for features that are derived from docgen capabilities (docgen, tab completion) use `TestArgumentContainer.java` as master input, since it contains all possible features. Updating that file with a new test feature (such as an arg that is a collection of enums) requires multiple expected output files to be updated.

PR #167, opened 2020-08-25
Update test output for a test that was added in the previous branch that was merged.

cmnbroad:
The previous branch (https://github.com/broadinstitute/barclay/pull/165, with WDL updates) added a new test, but the test output for that new test has to be updated to reflect the changes that were made to the input files in the last branch (https://github.com/broadinstitute/barclay/pull/160) so the tests on master will pass. (All of the WDL/Doc/Tab generation tests share the same test inputs - when changing a test input, often all of the test outputs need to be updated).


cmnbroad:
ty @lbergelson.

PR #169, opened 2020-09-14
Fix mutable HelpDoclet member declarations.

cmnbroad:
Make some static `HelpDoclet` members that were mutable into instance variables to prevent residual side effects from causing random test failures when tests are run serially.


droazen (approved):
:+1:

PR #174, opened 2020-10-08
Deduplicate argument names in help when fullname = shortname

lbergelson:
* When writing the help output don't output the argument in both short and long form if
  the shortname is the same as the long name.

  i.e.  --argument, -argument -> --argument


cmnbroad (commented):
Thanks @lbergelson !

PR #175, opened 2021-02-23
Change WDLGen default for localizationOptional to true.

cmnbroad:


droazen:
@cmnbroad Do you think it's safe to default to true here, or should we continue to require tool authors to manually label tool inputs/outputs as having cloud support?

cmnbroad:
Closing this in favor of leaving the default as is after more internal discussion.

PR #176, opened 2021-04-16
Update our artifactory password

lbergelson:
* Removing the artifactory password and username from the .travis.yml and updating them directly in travis settings instead
* This replaces our old invalid password with a new one and makes it easier to rotate in the future without requiring a new commit


droazen (approved):
:+1:

PR #187, opened 2022-10-04
Add @DeprecatedFeature annotation.

cmnbroad:
This also fixes an issue where positional args were previously not showing up in command line help.


cmnbroad:
@droazen I can't recall any reason why this was left as draft. I think its ready for review.

src/main/java/org/broadinstitute/barclay/argparser/PositionalArgumentDefinition.java
```diff
@@ -110,6 +110,60 @@ public void validateValues(final CommandLineArgumentParser commandLineArgumentPa
...
+    /**
+     * Return a string with the usage statement for this positional argument.
+     * @param argumentColumnWidth width reserved for argument name column display
+     * @param descriptionColumnWidth width reserved for argument description column display
+     * @return the usage string for this argument
+     */
+    public String getArgumentUsage(final int argumentColumnWidth, final int descriptionColumnWidth) {
+
+        final StringBuilder sb = new StringBuilder();
+        sb.append("--").append("POSITIONAL");
```
lbergelson, line 125: Should this include which position it should be in?  Is that easy to find?  I

src/main/java/org/broadinstitute/barclay/argparser/NamedArgumentDefinition.java
```diff
@@ -131,12 +128,6 @@ public NamedArgumentDefinition(
      */
     public boolean isAdvanced() { return getUnderlyingField().getAnnotation(Advanced.class) != null; }
 
-    /**
-     * return true if this argument has the {@code @Deprecated} annotation.
-     * @return true if this argument is advanced, otherwise false.
-     */
-    public boolean isDeprecated() { return getUnderlyingField().isAnnotationPresent(Deprecated.class); }
```
lbergelson, line 138: So previously this would be triggered by the standard java `Deprecated` annotation?  Should we continue to respect those as well?  I don't remember why exactly we decided to use a separate annotation instead of the default one.  So that intellij doesn't tag uses of it?

src/main/java/org/broadinstitute/barclay/argparser/PositionalArgumentDefinition.java
```diff
@@ -110,6 +110,60 @@ public void validateValues(final CommandLineArgumentParser commandLineArgumentPa
...
+        int labelLength = sb.toString().length();
+        int numSpaces = argumentColumnWidth - labelLength;
+        if (labelLength > argumentColumnWidth) {
+            sb.append("\n");
+            numSpaces = argumentColumnWidth;
+        }
+        Utils.printSpaces(sb, numSpaces);
+
+        String description = getArgumentDescription();
+        if (isDeprecated()) {
```
lbergelson, line 137: Should this little decorator be factored out so it isn't repeated?  Might make it more consistent if we change things in the future.

src/main/java/org/broadinstitute/barclay/help/GSONArgument.java
```diff
@@ -8,47 +8,41 @@
...
-        this.defaultValue = defaultValue;
-        this.minValue = minValue;
-        this.maxValue = maxValue;
-        this.minRecValue = minRecValue;
-        this.maxRecValue = maxRecValue;
-        this.kind = kind;
-        this.options = options;
+    @SuppressWarnings("unchecked")
+    public GSONArgument(final Map<String, Object> argMap) {
+        this.summary = argMap.get("summary").toString();
```
lbergelson, line 29: Moving this here seems like an improvement.

lbergelson (changes requested):
@cmnbroad A few minor comments.

src/main/java/org/broadinstitute/barclay/argparser/NamedArgumentDefinition.java
cmnbroad: The old code was carried over from the GATK3 implementation, but never fully embraced, and only worked for args IIRC (not tools or other documented features). I think we used a custom annotation both for the IntelliJ issue, and also because we wanted to be able to specify a reason/detail string, which the Java annotation doesn't have. There are a couple of usages of the old tag in GATK for arguments, but I'd prefer to update them rather than carry around (and test) both annotations.

src/main/java/org/broadinstitute/barclay/argparser/PositionalArgumentDefinition.java
cmnbroad: Done - factored out this decorator plus another method to format the description since that was also duplicated on two places.

src/main/java/org/broadinstitute/barclay/argparser/PositionalArgumentDefinition.java
cmnbroad: Not sure how to convey it, but positional args must always be first. But these are hardly ever used.

cmnbroad: I updated this String to say `"POSITIONAL (must be first)"`.

lbergelson (approved):
@cmnbroad Looks good, one nitpick abut the positionals still, feel free to merge and we cacn always address that later

src/main/java/org/broadinstitute/barclay/argparser/PositionalArgumentDefinition.java
```diff
@@ -110,6 +110,48 @@ public void validateValues(final CommandLineArgumentParser commandLineArgumentPa
...
+    /**
+     * Return a string with the usage statement for this positional argument.
+     * @param argumentColumnWidth width reserved for argument name column display
+     * @param descriptionColumnWidth width reserved for argument description column display
+     * @return the usage string for this argument
+     */
+    public String getArgumentUsage(final int argumentColumnWidth, final int descriptionColumnWidth) {
+
+        final StringBuilder sb = new StringBuilder();
+        sb.append("--").append("POSITIONAL (must be first)");
```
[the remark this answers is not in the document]
cmnbroad: @lbergelson There can only be one "positional" option per CLP. It can be a collection, so it can have multiple values, but they have to be first in the list. Its basically just an arg without a name. We should probably just get rid of it, but GATK and Picard each have one, I think.

cmnbroad:
Sorry @lbergelson, I had to add one more commit based on testing with GATK. I messed up the command line display formatting when I factored out display code in this PR, which shows up spectacularly in GATK. We were outputting this:
Screen Shot 2022-11-22 at 9 32 55 AM
instead of the output produced with this last commit:
Screen Shot 2022-11-22 at 9 33 29 AM

src/main/java/org/broadinstitute/barclay/argparser/PositionalArgumentDefinition.java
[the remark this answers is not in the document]
lbergelson: Huh, I thought we supported multiple positionals, sorry, ignore my comment then.

lbergelson (approved):

lbergelson:
@cmnbroad Good to merge.

cmnbroad:
Thanks @lbergelson!

PR #190, opened 2023-01-05
Fix reversed logic in the help template.

cmnbroad:
This template is only used for testing Barclay, but its still wrong.


lbergelson (approved):
👍 Good catch.

PR #191, opened 2023-01-31
Fix pom attributes, update readme to reflect new dependencies.

cmnbroad:
Fixes the pom so Sonatype will not reject Barclay because there are missing attributes. Also update the readme with the new dependencies.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -134,9 +134,9 @@ def basePomConfiguration = {
     }
 
     scm {
-        url 'scm:git@github.com:broadinstitute/barclay.git'
-        connection 'scm:git@github.com:broadinstitute/barclay.git'
-        developerConnection 'scm:git@github.com:broadinstitute/barclay.git'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -134,9 +134,9 @@ def basePomConfiguration = {
     }
 
     scm {
-        url 'scm:git@github.com:broadinstitute/barclay.git'
-        connection 'scm:git@github.com:broadinstitute/barclay.git'
-        developerConnection 'scm:git@github.com:broadinstitute/barclay.git'
+        url = 'scm:git@github.com:broadinstitute/barclay.git'
```
lbergelson, line 137: the mysteries of groovy....

lbergelson (approved):
@cmnbroad I feel like I was puzzled by this exact problem before and then forgot about it.

build.gradle
cmnbroad: Yes. It never ceases to entertain.

PR #193, opened 2023-02-08
Fix file leak in ClassFinder

lbergelson:
@cmnbroad I stumbled on this randomly and figured we should close it.

* Adds a try-with-resources block to close a zipfile in the class finder
* And the impossible to resist first java 17 related minor refactoring.


droazen (approved):
👍

PR #194, opened 2023-04-05
Adding new style versions to special arguments

lbergelson:
@droazen @cmnbroad Making the special arguments consistent but leaving the old way for now.


cmnbroad:
Other than that it looks good (although see https://github.com/broadinstitute/barclay/pull/123 for the last time we tried to do this).

PR #195, opened 2023-12-18
Update gradle to 8.5

lbergelson:
* gradle 8.5 is the first gradle that officially runs on java 21


cmnbroad (approved):
LGTM. Thanks for the upgrade @lbergelson.

PR #196, opened 2023-12-18
Replacing the gradle java plugin with java-library

lbergelson:
* updating vulnerable dependencies
* replacing use of commons-lang Path with new barclay Path
* reducing dependency scope where possible

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 plugins {
-    id "java"
+    id "java-library"
     id 'maven-publish'
     id 'signing'
     id 'jacoco'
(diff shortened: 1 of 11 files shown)


cmnbroad (commented):
Seems fine (we also use the old commons Pair class all over the place in GATK).

One request though - can you autogen/add `hashCode `/`equals` methods to `Pair`, since they get used all over the place.

cmnbroad (approved):

lbergelson:
@cmnbroad One of the nice things about records is they do that for you.  There are implicit equals/hashCode/toString methods that are reasonable and based on the declared fields.

https://docs.oracle.com/en/java/javase/17/language/records.html#GUID-6699E26F-4A9B-4393-A08B-1E47D4B2D263

lbergelson:
Getting rid of the commons pair in the interfaces is maybe a bit silly.  Might as well try to reduce the dependency surface a bit though.

cmnbroad:
Oh yeah - I knew that and completely forgot.

PR #197, opened 2023-12-18
Merging duplicate implementations of isConcrete

lbergelson:


cmnbroad (approved):
Oh, hm. I guess we only need one.

PR #198, opened 2023-12-18
lb update to 21

lbergelson:
- Updating to java 21
- update github actions to use java 21

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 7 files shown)


lbergelson:
@cmnbroad Java 21 is very aggressive with "this-escape" warnings.

lbergelson:
We probably don't want to merge this until we have similar branches for the whole pile of repos.

cmnbroad (approved):
Thanks @lbergelson !

cmnbroad (approved):
The new finals could just be replaced with "this-escape" annotations, but otherwise looks good.

